PR #1, opened 2020-08-06
Merge spike to get everything building

GMNGeoffrey:
Includes usage demonstrations with both http_archive and submodules.

The directory structure here is a bit weird because I'm deliberately
trying to emulate weird structures we might have to deal with.

The MLIR BUILD files are taken pretty much directly from the version
that TF exports. I restructured things a bit to avoid backward
dependencies on `@org_tensorflow` and commented out the cuda and vulkan
depencies for now to avoid having to set those up yet.

This gets everything building well enough to run
`bazel build @llvm-project//mlir/...` without failures and build all of
IREE with these build files. `bazel build @llvm-project//llvm/...` still
hits errors finding `.inc` files, but those errors exist with the TF
BUILD files. That is, we have parity with the TF files in terms of build
coverage.

The structure should make it much easier to expand and iterate though.
The entire overlay directory is fused into the llvm source directory
with symlinks, so we don't need to have an arbitrary limit of a few
files that push towards few BUILD files and are the reason for TF's
weird reverse dependencies.

My intent is to merge this with a rebase, so feedback on individual
commits (including their messages) is welcome.


llvm-bazel/overlay_directories.bzl
```diff
@@ -0,0 +1,78 @@
...
+    ).dirname
+    overlay_path = _join_path(
+        overlay_workspace_path,
+        repository_ctx.attr.overlay_path,
+    )
+
+    overlay_script = repository_ctx.path(
+        repository_ctx.attr._overlay_script,
+    )
+    python_bin = repository_ctx.which("python3")
```
stellaraccident, line 38: Note that windows python installations typically just have a "python" on the path. Might be worth falling back. It will produce odd errors if you happen to pick up a python2 unless if you do a check in the script itself.

llvm-bazel/overlay_directories.py
```diff
@@ -0,0 +1,83 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""Overlays two directories into a target directory using symlinks.
+
+Tries to minimize the number of symlinks created (that is, does not symlink
+every single file). Symlinks every file in the overlay directory. Only symlinks
+individual files in the source directory if their parent directory is also
+contained in the overlay directory tree.
+"""
+
```
stellaraccident, line 13: Might be worth adding:

```
from __future__ import print_function
```

And then including a little inline version check for python > 3:

```
import sys
if sys.version_info[0] < 3:
  raise RuntimeError("Must be invoked with a python 3 interpreter but was %s" % sys.executable)
```

That may be enough to parse and do an inline version check on python 2.7 without triggering other failures first. While we probably don't care much about python 2.7, it is still the default on a lot of supported OS's and if possible to be verbose about the mismatch, it can save a lot of time.

stellaraccident (approved):

llvm-bazel/overlay_directories.bzl
GMNGeoffrey: Done, thanks for the portability catch :-)

llvm-bazel/overlay_directories.py
GMNGeoffrey: I don't appear to even need the print_function import (even when the script contains a print usage). Invoking with python2 gives the expected error message like this.

stellaraccident: Oh right - without that Python2 will parse a print function as either a print statement with a parenthesized operand or a print statement of a tuple (if multiple operands). It will break with any named arguments, iirc. But since you just want the file to parse without error so it can execute the first couple of lines, it is fine either way.

chandlerc:
I'm still reading and learning how all this works, but generally looks nice!

PR #3, opened 2020-08-17
Switch http_archive demo to use tags

GMNGeoffrey:
We can use git tags to avoid having to look up commits corresponding to
a given LLVM revision. The tag format is `llvm-project-${LLVM_COMMIT}`.
I originally tried `llvm/llvm-project@${LLVM_COMMIT}` to make something
that is both autolinked by GitHub and unambiguous in the commit it
references but having URL-encoded characters seemed not worth it.

### DIFF
--- a/http-archive-demo/WORKSPACE
+++ b/http-archive-demo/WORKSPACE
@@ -6,15 +6,17 @@
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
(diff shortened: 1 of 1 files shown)


tpopp (approved):
LGTM. Just looking at it, I wouldn't be sure it works (specifically, where the strip_prefix is being used), but if you've tested it, then the idea looks fine.

PR #8, opened 2020-08-18
Document demos and move http-archive-demo to its own branch

GMNGeoffrey:
This isn't strictly necessary (like it is with the submodule-demo), but
having the same setup here makes sense. Because this has to reference
the commit it's using by SHA it's not actually possible to have it use
the current commit. I want to set up automation to make this always get
updated to match `main` but not pollute the main branch with lots of
automated commits.


README.md
```diff
@@ -15,17 +15,35 @@ suitable for more general usage.
 
 # Status
 
-Bazel builds can be run from the [llvm-bazel subdirectory](./llvm-bazel) for a
-build based on a submodule and [http-archive-demo](./http-archive-demo) for a
-build based on
+MLIR builds with Cuda and Vulkan support disabled.
+
+Building LLVM core has many issues with finding `.inc` files. These issues also
```
tpopp: My own opinion: I think it might be better to leave off comparisons to Tensorflow since it should be irrelevant once this is somewhat complete.

tpopp (approved):

README.md
GMNGeoffrey: Fair point. Done.

PR #10, opened 2020-08-19
Automatically integrate new LLVM commits

GMNGeoffrey:
When a change is pushed to `main`, it will be merged into
`auto-integrate` and then all LLVM commits after that point will get a
corresponding llvm-bazel commit tagged with `llvm-project-${COMMIT}`
that contains the corresponding build files.

Subsequent patches will make this trigger on new LLVM commits as well
(or on a cron) and try pulling in MLIR build file patches from TF.

The intended workflow is that we get candidates for all LLVM commits
out quickly. Users can with minimal delay reference a candidate set of
build files by using only the LLVM commit SHA. When we detect that some
LLVM commit requires a manual intervention, we push a new commit to
main with the submodule updated to that point. The automation then
creates new commits with corresponding tags for the following LLVM
commits (potentially overwriting the commit a tag pointed to). This
means that tags for commits on the auto-integrate branch are not
guaranteed to always point to the same commit and are therefore likely
not suitable for use in `http_archive` since the sha256 would be
different. Once a tag is applied to the main branch, we don't change it
however.

Tested:
You can see the results of this on the current `auto-integrate` branch.
I temporarily modified the workflow to also run on pushes to this PR
branch and it created a series of commits on the auto-integrate branch
with tags. I will drop these commits before merging this PR and let
them get generated from `main`.


scripts/clobber_merge.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
```
tpopp, line 6: Can you add a short description of what this script does?

scripts/generate_build_and_tag.sh
```diff
@@ -0,0 +1,31 @@
...
+set -e
+set -o pipefail
+
+SUBMODULE_DIR="third_party/llvm-project"
+
+LLVM_COMMIT="$(git submodule status -- ${SUBMODULE_DIR?} | awk '{print $1}' | tr -d '+')"
+
+sed -i '/Begin autogenerated content/Q' llvm-bazel/llvm-project-overlay/llvm/BUILD
+python3 ./llvm-bazel/generate_bazel_build.py \
+  --llvm_root=third_party/llvm-project/llvm \
```
tpopp: I think you want to use the SUBMODULE_DIR here.

GMNGeoffrey: Done, thanks for catching

scripts/clobber_merge.sh
GMNGeoffrey: Done. Missed this in my pass through to add documentation, thanks for catching :-)

tpopp (approved):

PR #11, opened 2020-08-27
WIP: Begin filling in a hand-written port of LLVM to Bazel.

chandlerc:
This replaces the automatically generated BUILD file with rules directly
written against the LLVM project. Rules are designed to use narrow globs
to be reasonably resilient to simple source additions and removals.
Headers and layering are strictly enforced here. There is significant
logic to build the targets, with some libraries being merged due to
`#include` cycles that cannot be split.

Not all of the LLVM libraries are included yet, just enough to build the
`opt` and `llc` tools, including most of the targets, and the testonly
libraries and tools that include tricky parts of GoogleTest and
GoogleMock.

Still lots of cleanup to do, and then expanding this to cover the entire
LLVM library collection.


chandlerc:
This now has every `cc_library` and `cc_binary` rule that I think makes sense in the main `BUILD` file. The MLIR `BUILD` file builds against this cleanly. I think this is worth reviewing.

Outstanding things that may make sense for this PR:
1) Testing on non-Linux platforms. I've been doing very basic testing here, there may be problems on other platforms. Would appreciate help here.
2) Any refactoring folks would like of the overall structure.

One thing that may or may not make sense here: removing some of the starlark code that is no longer used. I'm happy to do that here or in a follow-up.

Things that I think make sense in subsequent PRs:
- Add unittest BUILD files.
- Add some basic sh_tests for the binaries here.
- Add a Clang BUILD file. (This will take a while...)

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -30,2213 +14,1351 @@ exports_files(["LICENSE.TXT"])
...
 
+# FIXME: Would be nice to have a select here on the compiler so that we can use
+# a more minimal list for Clang than we need for GCC.
+llvm_copts = [
+    "$(STACK_FRAME_UNLIMITED)",
+    "-Wno-unused-parameter",
+    "-Wno-comment",
+    "-Wno-class-memaccess",
+    "-Wno-maybe-uninitialized",
+    "-Wno-misleading-indentation",
```
GMNGeoffrey, line 29: I'm getting oodles of warnings here. If we can't set up a select now, I'd prefer to either move these to a bazelrc config option or also set `-Wno-unknown-warning-option`, although maybe that upsets gcc.

> warning: unknown warning option '-Wno-class-memaccess'; did you mean '-Wno-class-varargs'? [-Wunknown-warning-option]
> warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
> warning: unknown warning option '-Wno-misleading-indentation'; did you mean '-Wno-binding-in-condition'? [-Wunknown-warning-option]

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -30,2213 +14,1351 @@ exports_files(["LICENSE.TXT"])
...
-    ],
-    defines = llvm_defines,
-    includes = ["include"],
-)
-
-# A creator of an empty file include/llvm/Support/VCSRevision.h.
-# This is usually populated by the upstream build infrastructure, but in this
-# case we leave it blank. See upstream revision r300160.
+# TODO: Need to replace this with something that actually extracts the git
+# commit from the LLVM submodule.
```
GMNGeoffrey: We should make this generic/substitutable so that users don't have to include this or LLVM as a submodule (e.g. could use http_archive).

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -30,2213 +14,1351 @@ exports_files(["LICENSE.TXT"])
...
-)
+# These rules substitute fairly generic Linux and x86 flavored configs
+# currently. We could extend it in-place with selects-style switching on
+# platforms, or by extracting parts of the dictionary to be computed in with
+# a separate starlark function where the config switching can be managed.
+#
+# TODO: Priority list of enhancements needed here:
+# - Reflect `opt`, `dbg`, etc. features.
+# - Basic port to Mac for convenience.
+# - Extract into something cleaner based on Starlark.
```
GMNGeoffrey, line 128: Oof I feel like "extract into something cleaner" has got to be a higher level goal here. Could just be a macro. It's actually not clear to me that deriving and substituting from the llvm-config.h.cmake file is better than just handwriting this file? I think that would be easier. The overlay is generic, so we don't have to have everything in a build file.

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -30,2213 +14,1351 @@ exports_files(["LICENSE.TXT"])
...
-        ":MCDisassembler",
-        ":Support",
+        ":Demangle",
         ":config",
     ],
 )
 
 cc_library(
-    name = "AVRInfo",
+    name = "LineEditor",
```
GMNGeoffrey, line 659: I see this error:

```
ERROR: /usr/local/google/home/gcmn/.cache/bazel/_bazel_gcmn/a584c7045e7d98d45584612e9b40b482/external/llvm-project/llvm/BUILD:652:11: C++ compilation of rule '@llvm-project//llvm:LineEditor' failed (Exit 1) clang failed: error executing command /usr/bin/clang -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -fcolor-diagnostics -fno-omit-frame-pointer -g0 -O2 '-D_FORTIFY_SOURCE=1' -DNDEBUG -ffunction-sections ... (remaining 31 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
warning: unknown warning option '-Wno-class-memaccess'; did you mean '-Wno-class-varargs'? [-Wunknown-warning-option]
warning: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'? [-Wunknown-warning-option]
warning: unknown warning option '-Wno-misleading-indentation'; did you mean '-Wno-binding-in-condition'? [-Wunknown-warning-option]
external/llvm-project/llvm/lib/LineEditor/LineEditor.cpp:18:10: fatal error: 'histedit.h' file not found
#include <histedit.h>
```

Needed to install libedit-dev. I'm surprised Bazel is using system dependencies like that

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -30,2213 +14,1351 @@ exports_files(["LICENSE.TXT"])
...
-    hdrs = glob([
-        "include/llvm/Target/AVR/Disassembler/*.h",
-        "include/llvm/Target/AVR/Disassembler/*.def",
-        "include/llvm/Target/AVR/Disassembler/*.inc",
-        "lib/Target/AVR/Disassembler/*.h",
+    textual_hdrs = glob([
+        "include/llvm/Support/*.def",
     ]),
-    copts = llvm_copts + ["-Iexternal/llvm-project/llvm/lib/Target/AVR"],
+    visibility = ["//visibility:public"],
```
GMNGeoffrey: This is the default visibility, so it doesn't make much sense to have it here, I think. I've been thinking some about how to make Bazel visibility make sense in a multi-repo world (it works much better in the monorepo). For now, I think drop the default visibility from the package or drop all the explicit public visibility here.

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -2245,2491 +1367,2744 @@ cc_library(
...
         ":MC",
-        ":MCDisassembler",
         ":Support",
         ":config",
     ],
 )
 
+llvm_target_list = [
+    {
+        "name": "AArch64",
```
GMNGeoffrey, line 1759: Do "name" and "dir_name" ever differ? Is it useful having both?

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -2245,2491 +1367,2744 @@ cc_library(
...
+            "utils/unittest/googletest/include/gtest/gtest_pred_impl.h",
+        ],
+    ) + [
+    ],
+    hdrs = ["utils/unittest/googletest/include/gtest/gtest.h"],
+    # TODO: the hard-coding of `external/llvm-project` path components here is
+    # really unfortunate. It is essentially baking in an implementation detail
+    # of how Bazel runs the compiler when compiling LLVM in the exact
+    # configuration we use.
+    copts = llvm_copts + ["-Iexternal/llvm-project/llvm/utils/unittest/googletest"],
```
GMNGeoffrey, line 3702: Yeah this is on my TODO list as well

GMNGeoffrey:
I can build `@llvm-project//...` :tada:

> 1. Testing on non-Linux platforms. I've been doing very basic testing here, there may be problems on other platforms. Would appreciate help here.

I've been leaving that for later. TF had some selects by platform that I took out because they relied on tf targets. Need to figure out the project-independent way of doing that.

> 2. Any refactoring folks would like of the overall structure.
> 
> One thing that may or may not make sense here: removing some of the starlark code that is no longer used. I'm happy to do that here or in a follow-up.

I'd say delete it now if it's unused, but also fine with a follow-up.

GMNGeoffrey (commented):
Woooo! This is great. I think the high-level question is how maintainable this all is for the non-Chandler's of the world. I think it's better than the partial-autogen script, but we should probably look at at least tool-assisted updating.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: There's no real way to suppress these with GCC.

But you should only see them when some other warning is emitted. I think the only one of these left is for C code being built with these flags which include a C++-specific warning. I'm going to try to fix that upstream so that I can just remove the flag and this will be clean again.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: Yep.

Do you think that's OK to do as a follow-up? If it needs to happen now, I can start trying to learn how that kind of stuff works in Bazel, but ideally I'd like to just fix this afterward.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: I actually originally had LLVM using a hand written file inside Google, and it was a bit cleaner and worked for years.

But then we found a pretty terrible down side: there was nothing to catch when it drifted apart from upstream.

The huge advantage of this approach is that it forces you to deal with any additional configuration knobs added upstream -- they break the build. I think that's worth the overhead of having this inline.

Still, this point was supposed to be trivial or lost. I think the best way to do this is to just sink this down into starlark. I actually think the somewhat verbose substitution approach works pretty nicely in practice. I actually wrot e the first version of this long enough ago that I've even had to update it a couple of times and its surprisingly easy. You go into the .cmake file, extract the new lines, put them here, and then adjust to reflect the desired substitution.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: I always forget that Bazel's default is public, sorry.

Dropped all the redundant public visibility lines.

Happy to see a real visibility approach here eventually.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: The goal is genuinely to link against the system library here at an ABI level (rather than any kind of source dependency).

I can disable this one (and I've made that edit) because we have a fallback. But there are other places where we depend on fundamental system libraries. For example the `Option` library uses `libdl.so` to get the executable name. There are a few other places. The goal of these is very much to connect to the installed *system* libraries and not to use any kind of packaged library.

Is there a better way to handle these? A way to write some early test that the necessary system libraries are installed?

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: No! Yay! Fixing!

chandlerc (commented):
Regarding maintenance: I'm not sure it'll really be an issue.

We've been manually updating this BUILD file structure for years on a daily basis. It isn't trivial, but it is a very minor burden. The incremental updates are pretty straight forward.

And most of these libraries are really boring. The target libraries are tricky of course, but those aren't really changing.

Last but not least, we should be able to unify the updates to our internal copy and this eventually?

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: I see them every build step unto infinity :-D What about moving these to the bazelrc then where it can be compiler dependent? (Or just fixing them upstream works :grin:)

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Yes for sure this is fine as a follow-up. I just wanted to call out that it wasn't necessarily a submodule.

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Ah good point. And I agree that doing it with a script wrapped in starlark. Looks like gn has one. Maybe we could even use the same one: https://github.com/llvm/llvm-project/blob/master/llvm/utils/gn/build/write_cmake_config.py

That can be a follow-up though :-)

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: We could probably rig something up with repo rules. I think I'd have to understand better what we're testing and what we're guarding against. I'm not too worried, I was just surprised that this seemed to escape the sandbox.

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Bazel's default is private, but this package is declared as default public (in the package call up at the top). We can change that if we want

GMNGeoffrey:
> Regarding maintenance: I'm not sure it'll really be an issue.
> 
> We've been manually updating this BUILD file structure for years on a daily basis. It isn't trivial, but it is a very minor burden. The incremental updates are pretty straight forward.
> 
> And most of these libraries are really boring. The target libraries are tricky of course, but those aren't really changing.
> 
> Last but not least, we should be able to unify the updates to our internal copy and this eventually?

Yes all of these things are true. Part of my concern is that until we make this more widely supported, maintainers of this repo (AKA me right now :grin:) are going to need to keep this up to date. Let's try it out.

GMNGeoffrey:
I think llvm-bazel/llvm-project-overlay/llvm/expand_cmake_vars.py is also going to be unused now. Can you delete it as part of this PR?

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -30,2213 +14,1433 @@ exports_files(["LICENSE.TXT"])
...
-        ":AVRDesc",
-        ":AVRInfo",
-        ":MC",
-        ":MCParser",
         ":Support",
         ":config",
     ],
 )
 
+# exports_files(["include/llvm/Option/OptParser.td"])
```
GMNGeoffrey: Delete?

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Ah we actually run into the same issue with zlib. The build now fails on IREE's remote executors building `Support` complaining about zlib.h I assume because they don't have zlib.

Can you add the "@zlib" dependency back to `Support`? Or should this just be an installed dependency? https://reviews.llvm.org/D79219 also has some discussion about detecting zlib in cmake.

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -30,2213 +14,1433 @@ exports_files(["LICENSE.TXT"])
...
+        "include/llvm-c/Error.h",
+        "include/llvm-c/ExternC.h",
+        "include/llvm-c/ErrorHandling.h",
+        "include/llvm-c/Support.h",
+        "include/llvm-c/Types.h",
+    ],
+    copts = llvm_copts,
+    includes = ["include"],
+    linkopts = [
+        "-lz",
```
GMNGeoffrey: I now get errors compiling IREE with these build files (was clean before), even after restoring the zlib dependency:

> ERROR: [snip]/.cache/bazel/_bazel_gcmn/c17d50ae90e3b490378de262a2ee1611/external/llvm-project/mlir/BUILD:3144:10: Linking of rule '@llvm-project//mlir:mlir-tblgen' failed (Exit 1)
/usr/bin/ld.gold: error: cannot find -lz
/usr/bin/ld.gold: error: cannot find -ltinfo
bazel-out/host/bin/external/llvm-project/llvm/_objs/Support/Process.o:Process.cpp:function llvm::sys::Process::FileDescriptorHasColors(int): error: undefined reference to 'setupterm'
bazel-out/host/bin/external/llvm-project/llvm/_objs/Support/Process.o:Process.cpp:function llvm::sys::Process::FileDescriptorHasColors(int): error: undefined reference to 'tigetnum'
bazel-out/host/bin/external/llvm-project/llvm/_objs/Support/Process.o:Process.cpp:function llvm::sys::Process::FileDescriptorHasColors(int): error: undefined reference to 'set_curterm'
bazel-out/host/bin/external/llvm-project/llvm/_objs/Support/Process.o:Process.cpp:function llvm::sys::Process::FileDescriptorHasColors(int): error: undefined reference to 'del_curterm'

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Ah I just realized that expand_cmake_vars is a script we already have that does this. It might also help to address the issues I'm hitting with terminfo and such with IREE.

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Here are my patches to make IREE build, a property I'm interested in preserving ;-P

https://github.com/google/llvm-bazel/commit/2af8cba83fb593ddb76d4a2c1ac496e4d11f5ee5

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: I see, sorry. The issue there is using Clang instead of GCC. I've updated this to hopefully be tolerable for both GCC and Clang.

I'm happy to work on a better model here, but would prefer to do that in follow-ups. I would like to avoid a top-level entry as users of this may not want to disable these warnings on their own code.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: Clarified the comment.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: All these scripts do is handle the substitution. That's not really the interesting part here -- Bazel already has a built in way of doing that via Starlark that we're using here and I'd much rather keep using that. Having to format the substitutions the right way seems like a minor inconvenience compared to being able to do this entirely in Bazel itself.

The trickier thing is to use `select`s to actually *configure* the substitutions in a way that match different build environments. That isn't helped by either of these scripts really -- we'd still need that logic regardless of the mechanism used for the textual substitution.

My big question is: do you want me to start factoring this into a separate starlark bit in this PR, or in follow-ups?

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: Ok, I'm just disabling both zlib and terminfo for now. That way the zlib http_archive dependency can be kept out too for now.

We can figure out if/how to restore these in follow-ups.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: See above -- I've disabled zlib so that it shouldn't be necessary. Let me know if that works? (If it doesn't can add it back of course.)

chandlerc (commented):
PTAL, I think most of the most important stuff is in place.

chandlerc:
FYI: I did a manual sync of this to top-of-tree LLVM. You can see the commit needed to get everything to build here:
https://github.com/google/llvm-bazel/commit/80ab78b88c3480a9806af4e903fea419d1db1474

The delta is about 2 weeks of churn upstream. Most of the churn in MLIR, and most of that I was able to update by just grabbing fresh code from TF. Only a couple of further edits needed beyond that. It was only a little tricky to merge because there are still some (understandable) differences.

If/when this and #12 land, I can turn it into a PR of course, mostly just wanted you to be able to see what it looked like to update config.h.cmake and some libraries. It's not too bad.

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Yeah good point. We can do this entirely in starlark and have a nicely formatted substitution actually. Configuring based on selects also sounds good.

Follow-up sounds good for both

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: What we do in IREE is use per_file_copt to [disable diagnostics on code we don't control and set `-Werror` on code we do](https://github.com/google/iree/blob/dcef2858374139821a7dbe2f5177706211438c71/build_tools/bazel/iree.bazelrc#L67-L70).

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: I still need an additional patch to enable dlopen, which we use

https://github.com/google/llvm-bazel/compare/chandlerc-hand-port...iree-patches-to-hand-port-2

GMNGeoffrey:
> FYI: I did a manual sync of this to top-of-tree LLVM. You can see the commit needed to get everything to build here:
> [80ab78b](https://github.com/google/llvm-bazel/commit/80ab78b88c3480a9806af4e903fea419d1db1474)
> 
> The delta is about 2 weeks of churn upstream. Most of the churn in MLIR, and most of that I was able to update by just grabbing fresh code from TF. Only a couple of further edits needed beyond that. It was only a little tricky to merge because there are still some (understandable) differences.

Nice. One of the next things I'm going to look at is exporting the internal files directly (to a separate branch) rather than through TF wonkiness.

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Ok I think I don't understand well enough what these things are being used for to understand why it *needs* to be a system lib, but we can discuss later :-)

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: I'd like for the flags needed for LLVM to be provided by the LLVM BUILD file though... We can give up at some point, but I'm happy to keep iterating here for now, as long as the current state works for IREE?

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: Ah, yes. That seems like a good change. Suggest it in this PR? Or merge that one?

GMNGeoffrey: Done.

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Yeah agreed. And this should work fine for IREE. Thanks :-)

GMNGeoffrey (approved):

PR #12, opened 2020-08-31
Remove build file generation

GMNGeoffrey:
With https://github.com/google/llvm-bazel/pull/11, we will use
handwritten Bazel BUILD files.


chandlerc (approved):
LGTM (once #11 lands of course)

GMNGeoffrey:
I actually want to merge this one before #11 because I currently have things set up to traverse LLVM commits and run the generation script, tagging new commits on this repo. Don't want that to happen when #11 is merged. It will append the generated build file to your handwritten one

PR #16, opened 2020-09-13
Remove references to repo name by alias

GMNGeoffrey:
This is a proper repository that shouldn't need to refer to itself by
an alias (which may be different in projects that depend on it).

Tested:
`bazel build @llvm-project//...`


tpopp (approved):

tpopp:
Let me know if you would like me to "auto-submit" changes like this also.

GMNGeoffrey:
> Let me know if you would like me to "auto-submit" changes like this also.

Yeah I wish there was an easier equivalent on GitHub. One of the unfortunate things is that there are still a lot of decisions to be made with submission (merge strategy, commit description, etc.), which I think make it a bit harder to leave it to reviewers. In IREE, I've tried to establish conventions for how to do this, like PR description == commit description, so there's less ambiguity.

PR #17, opened 2020-09-13
Update status in README

GMNGeoffrey:


README.md
```diff
@@ -15,9 +15,14 @@ suitable for more general usage.
 
 # Status
 
-MLIR builds with Cuda and Vulkan support disabled.
+MLIR has Cuda and Vulkan support disabled.
 
-Building LLVM core has many issues with finding `.inc` files.
+The whole project builds, e.g.
```
tpopp: Nit: "The whole project builds" sounds like it's surprising that happens, while it should be generally expected (of any non-new project). I'd recommend something more like "The project can be  build with a command such as" or "To build, run:" or similar.

tpopp (approved):

README.md
GMNGeoffrey: Well the status update is that now everything builds (previously that wasn't the case). And in contrast we have no tests (yet). Since this whole project is just "add build files for LLVM", I think it's a little different. Normally when developing a project, you'd have some build system as it was growing (how else were you developing). In this case we're trying to adopt a new build system for an existing massive project.

GMNGeoffrey: I do agree that sentence sounds weird though. I've reworded.

PR #18, opened 2020-09-13
Remove zlib build configuration

GMNGeoffrey:
We've disabled zlib for now. Eventually, we want to enable based on its
availability as a system library and link against that at the ABI level.

Tested:
`bazel build @llvm-project//...`


chandlerc (approved):
LGTM

PR #19, opened 2020-09-13
Fix auto-integrate action

GMNGeoffrey:
Unconditionally commit submodule updates to the auto-integrate branch.
Having a check for a clean working directory there makes no sense as we
will always have a change to the submodule.

Separately, have an action for just tagging the HEAD commit on `main`.

### DIFF
--- a/.github/workflows/auto_tag.yml
+++ b/.github/workflows/auto_tag.yml
@@ -1,0 +1,31 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Tags the HEAD commit on main with the corresponding LLVM commit
+
+name: Auto Tag
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  auto_integrate:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checking out repository
+        uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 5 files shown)


scripts/tag_rev.sh
```diff
@@ -1,27 +1,21 @@
-#!/bin/bash
+a#!/bin/bash
```
tpopp: Typo

.github/workflows/auto_tag.yml
```diff
@@ -0,0 +1,31 @@
...
+  auto_integrate:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checking out repository
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Tagging head commit
+        run: ./scripts/tag_rev.sh
+      - name: Fetching integration branch
```
tpopp: Is there a missing line? I was just surprised to see 2 "name" lines without 2 commands corresponding to them.

tpopp (changes requested):

tpopp:
I'm blocking because of the `#!` -> `a#!` but I don't actually have any issues after.

.github/workflows/auto_tag.yml
GMNGeoffrey: This line should have been deleted

GMNGeoffrey:
> I'm blocking because of the `#!` -> `a#!` but I don't actually have any issues after.

Based on your comment, I'm going to merge this now so we don't have to wait for another round of timezones :-)

PR #20, opened 2020-09-14
Integrate LLVM at llvm/llvm-project@c799f873cb9f

GMNGeoffrey:
Cherry-picked changes to the Google internal MLIR BUILD files (now
exported to the `google-mlir` branch). We could do some sort of merging
strategy, but for now this just avoids having to grab them in a
roundabout fashion from TF (where they also have some TF-specific
weirdnesses attached).

### DIFF
--- a/llvm-bazel/llvm-project-overlay/mlir/BUILD
+++ b/llvm-bazel/llvm-project-overlay/mlir/BUILD
@@ -1,3 +1,7 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
 # Description:
 #   The MLIR "Multi-Level Intermediate Representation" Compiler Infrastructure
 
--- a/llvm-bazel/llvm-project-overlay/mlir/BUILD
+++ b/llvm-bazel/llvm-project-overlay/mlir/BUILD
@@ -1,21 +1,16 @@
-# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
-# See https://llvm.org/LICENSE.txt for license information.
-# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-
 # Description:
 #   The MLIR "Multi-Level Intermediate Representation" Compiler Infrastructure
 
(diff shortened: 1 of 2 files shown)


llvm-bazel/llvm-project-overlay/mlir/BUILD
```diff
@@ -1,12 +1,8 @@
-# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
-# See https://llvm.org/LICENSE.txt for license information.
-# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
chandlerc: Why are we losing this in the file? Just seems inconsistent... But maybe I don't understand the context here.

GMNGeoffrey: I've yet to really get a clear answer on what license headers are actually supposed to *do* and where they're required and how hard we should try to put them there: e.g. all of TF doesn't put license headers on its build files. Similarly, LLVM doesn't put license headers on its CMakeLists.txt files. In this case, the internal file doesn't have a header and it's exported to TF as well, so I can't really put an LLVM header on it. I could add the header on export, but it didn't seem worth the trouble.

For consistency, we could drop the header from the llvm subproject build files as well for now.

chandlerc: I think we should add the header if we want this to eventually be viable for inclusion in the larger LLVM project FWIW.

I don't have any strong opinions about whether to do so in the TF project. But unless it is a significant burden, I think having the headers is useful documentation to have at the top of each file and I'd prefer to be consistent about it.

GMNGeoffrey: Ok I added back the headers and I'll look at adding them on export. It shouldn't be too bad here because the header doesn't include the year, which is supposed to be the oldest creation date, so can't just be added as the current date each time.

I'm surprised you think this would cause issues for inclusion in the LLVM project though. The whole repo is licensed under the Apache-LLVM license and these headers are just reminders AFAIU. There aren't license headers in the CMakeLists.txt files in the llvm-project, which seems like the closest analogue.

chandlerc (commented):
LGTM, thanks!

chandlerc (approved):
(this time checking the box)

PR #21, opened 2020-09-16
Remove a disabled GCC warning now that LLVM doesn't hit it.

chandlerc:
LLVM's code is fixed in https://reviews.llvm.org/D87755

Once that lands, this warning doesn't need to be disabled here. This is
particularly nice because it stops GCC from complaining when we try to
disable it in C compilations.


chandlerc:
Not working on this any more as the warnings have been moved to a less problematic place.

PR #23, opened 2020-09-17
Enable vulkan support

GMNGeoffrey:
Mostly adopts the configuration used in https://github.com/google/iree.

The specific `WORKSPACE` configuration of this (as in all things) is
just one example of how to do this. The key thing is that in order to
use the VulkanRuntime target, users must create appropriate
`@vulkan_headers` and `@vulkan_sdk` repo rules. I'm not aware of an
easy way to make this configurable (not have it rely on a specific
name).

We leave all the targets that rely on external dependencies as "manual"
so that users do not need to configure any additional dependencies to
use llvm-bazel.

TESTED=`bazel build @llvm-project//...`

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,8 +15,8 @@ suitable for more general usage.
 
 # Status
 
-MLIR has Cuda and Vulkan support disabled. LLVM does not yet have configuration
-detection and hardcodes values in the config.
+MLIR has Cuda support and the mlir-vulkan-runner disabled. LLVM does not yet
+have configuration detection and hardcodes values in the config.
 
 # Usage
 
--- a/llvm-bazel/WORKSPACE
+++ b/llvm-bazel/WORKSPACE
@@ -3,9 +3,29 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 load(":configure.bzl", "llvm_configure")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 
 llvm_configure(
     name = "llvm-project",
     overlay_path = "llvm-project-overlay",
     src_path = "../third_party/llvm-project",
+)
+
+maybe(
+    http_archive,
+    name = "vulkan_headers",
+    strip_prefix = "Vulkan-Headers-9bd3f561bcee3f01d22912de10bb07ce4e23d378",
+    sha256 = "19f491784ef0bc73caff877d11c96a48b946b5a1c805079d9006e3fbaa5c1895",
+    urls = [
+        "https://github.com/KhronosGroup/Vulkan-Headers/archive/9bd3f561bcee3f01d22912de10bb07ce4e23d378.tar.gz",
+    ],
+    build_file = "//third_party_build:vulkan_headers.BUILD",
+)
+
+load(":vulkan_sdk.bzl", "vulkan_sdk_setup")
+
+maybe(
+    vulkan_sdk_setup,
+    name = "vulkan_sdk",
 )
--- a/llvm-bazel/llvm-project-overlay/mlir/BUILD
+++ b/llvm-bazel/llvm-project-overlay/mlir/BUILD
@@ -3129,6 +3129,9 @@ cc_library(
     hdrs = [
         "tools/mlir-vulkan-runner/VulkanRuntime.h",
     ],
+    tags = [
+        "manual",  # Requires an external dependency.
+    ],
     deps = [
         ":IR",
         ":Pass",
@@ -3138,7 +3141,7 @@ cc_library(
         ":Support",
         "//llvm:Support",
         "@vulkan_headers",
-        "@vulkan_sdk//:sdk"
+        "@vulkan_sdk//:sdk",
     ],
 )
 
@@ -3146,6 +3149,9 @@ cc_binary(
     name = "tools/libvulkan-runtime-wrappers.so",
     srcs = ["tools/mlir-vulkan-runner/vulkan-runtime-wrappers.cpp"],
     linkshared = True,
+    tags = [
+        "manual",  # Requires an external dependency (via VulkanRuntime)
+    ],
     deps = [
         ":VulkanRuntime",
         "//llvm:Support",
--- a/llvm-bazel/llvm-project-overlay/mlir/BUILD
+++ b/llvm-bazel/llvm-project-overlay/mlir/BUILD
@@ -3129,13 +3129,6 @@ cc_library(
     hdrs = [
         "tools/mlir-vulkan-runner/VulkanRuntime.h",
     ],
-    data = [
-        "//third_party/swiftshader/src:libvk_swiftshader.so",
-        "//third_party/swiftshader/src:vulkan_icd_manifest_files",
-    ],
-    tags = [
-        "manual",  # TODO(gcmn) enable this target
-    ],
     deps = [
         ":IR",
         ":Pass",
@@ -3144,8 +3137,8 @@ cc_library(
         ":StandardOps",
         ":Support",
         "//llvm:Support",
-        "//third_party/vulkan_headers",
-        "//third_party/vulkan_loader",
+        "@vulkan_headers",
+        "@vulkan_sdk//:sdk"
     ],
 )
 
@@ -3153,9 +3146,6 @@ cc_binary(
     name = "tools/libvulkan-runtime-wrappers.so",
     srcs = ["tools/mlir-vulkan-runner/vulkan-runtime-wrappers.cpp"],
     linkshared = True,
-    tags = [
-        "manual",  # TODO(gcmn) enable this target
-    ],
     deps = [
         ":VulkanRuntime",
         "//llvm:Support",
--- a/llvm-bazel/third_party_build/vulkan_headers.BUILD
+++ b/llvm-bazel/third_party_build/vulkan_headers.BUILD
@@ -1,0 +1,27 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+package(default_visibility = ["//visibility:public"])
+
+# Exports all headers but defining VK_NO_PROTOTYPES to disable the
+# inclusion of C function prototypes. Useful if dynamically loading
+# all symbols via dlopen/etc.
+# Not all headers are hermetic, so they are just included as textual
+# headers to disable additional validation.
+cc_library(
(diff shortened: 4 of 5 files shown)


README.md
```diff
@@ -15,8 +15,8 @@ suitable for more general usage.
 
 # Status
 
-MLIR has Cuda and Vulkan support disabled. LLVM does not yet have configuration
-detection and hardcodes values in the config.
+MLIR has Cuda support and the mlir-vulkan-runner disabled. LLVM does not yet
+have configuration detection and hardcodes values in the config.
```
chandlerc, line 19: It isn't clear to me if this is already the case, but can the `vulkan_sdk_setup` routine be used by consumers of the repo? If not, could you rearrange things to make that work?

Then, maybe it makes sense to document how users interested in enabling Vulkan rules can use your repository rule from their `WORKSPACE` file (unless they want to write their own)?

llvm-bazel/WORKSPACE
```diff
@@ -3,9 +3,29 @@
...
+maybe(
+    http_archive,
+    name = "vulkan_headers",
+    strip_prefix = "Vulkan-Headers-9bd3f561bcee3f01d22912de10bb07ce4e23d378",
+    sha256 = "19f491784ef0bc73caff877d11c96a48b946b5a1c805079d9006e3fbaa5c1895",
+    urls = [
+        "https://github.com/KhronosGroup/Vulkan-Headers/archive/9bd3f561bcee3f01d22912de10bb07ce4e23d378.tar.gz",
+    ],
+    build_file = "//third_party_build:vulkan_headers.BUILD",
+)
```
chandlerc, line 24: Is the `maybe` here actually necessary?

It seems better to condition this on the Vulkan SDK being installed? Or does the `maybe` enable that?

If Bazel makes the above hard, just let me know.

Perhaps more interestingly -- could you sink this into the Vulkan SDK repository rule? Or into an adjacent one so that users that want Vulkan support don't have to replicate this rule to get the behavior?

llvm-bazel/llvm-project-overlay/mlir/BUILD
```diff
@@ -3144,18 +3137,15 @@ cc_library(
         ":StandardOps",
         ":Support",
         "//llvm:Support",
-        "//third_party/vulkan_headers",
-        "//third_party/vulkan_loader",
+        "@vulkan_headers",
+        "@vulkan_sdk//:sdk"
```
chandlerc: Can you create something that is `select`-able on, and make the entire rule (and the dependencies on the rule below) conditioned on Vulkan SDK being available? Something like: https://stackoverflow.com/questions/63677527/conditionally-create-a-bazel-rule-based-on-config

As it is, if you touch this BUILD file (even to query) without the SDK, it'll fail. =/ This seems especially bad for users of LLVM that aren't *specifically* interested in MLIR but now a pattern that just happens to include the directory will cause build errors. Unless I'm missing something?

chandlerc (changes requested):

README.md
GMNGeoffrey: Yes it should be. I'm intending to provide a demo of that in our demos. But basically they just copy what I've got in the WORKSPACE.

llvm-bazel/WORKSPACE
GMNGeoffrey: The repo rule itself conditions on the Vulkan SDK being installed and creates an empty rule (pushing this to a link-time error) if not. So you have to create the "vulkan_sdk" repo, but you don't actually have to have the Vulkan SDK.

The `maybe` here doesn't actually do much. `maybe` just wraps a repo rule and only creates the repo if the name doesn't already exist (without `maybe` this would be an error). This creates a "first one wins" approach to repo aliases, which isn't ideal, but is better than the situation you get without `maybe` where it becomes impossible to configure a repository (e.g. TF) without getting its dependencies along for the ride. You'll see a few places in IREE where we have to do really weird stuff because of this (TF's setup predates the introduction of maybe).

In this case, it doesn't matter really because I'm not using anyone else's configuration rules. The reason I'm using it is because I expect whatever I write here will get copy-pasta'd all over the place so I want to demo best practices.

llvm-bazel/llvm-project-overlay/mlir/BUILD
GMNGeoffrey: That's not quite right. To query this rule, you don't even need `@vulkan_sdk` defined. To build (or cquery) this rule, you need it defined, but it doesn't actually have to do anything. To run the things that depend on this rule, you need to actually have the Vulkan SDK or you'll get link errors. A nice feature of that is that you can build things even without having the Vulkan SDK. Does that seem acceptable? Or do you want this rule to not even be defined if the vulkan_sdk isn't available. Or we could have it select to only add the vulkan_sdk if it's available, which would have the same feature of being able to build without it. I can look into that as well.

Another option would be to leave this rule (and any that depends on an external repo alias as `manual`). Then they won't get picked up in wildcard builds so no one should care and they don't have to have `@vulkan_sdk` defined at all. The disadvantage is that `bazel build @llvm-project//...` no longer actually builds everything, we'd need to pipe query to build/test, although piping query to test is actually the only way to get bazel to build and test everything in a single command (just `bazel test` won't build binaries unused in tests). In that case we'll need to find another way to tag the rules currently tagged manual to differentiate them as not buildable.

GMNGeoffrey:
So this might come to platform detection again and we should use a similar strategy to whatever we do for zlib. The bazely way would usually be to make the entire command line specific. Instead of using bazel repo rules, as we previously discussed, we could make this something that gets put in a env-specific bazelrc. In IREE we've got a configured.bazelrc that you generate by [running a script](https://github.com/google/iree/blob/main/configure_bazel.py)

llvm-bazel/WORKSPACE
chandlerc: Sure, but my goal here is to not fetch the headers unless the SDK is actually available.

Basically -- if the SDK is available, you get the headers and the SDK. If not, you get neither.

llvm-bazel/llvm-project-overlay/mlir/BUILD
chandlerc: Ideally, we'd disable this rule (either by marking it manual or something else) and disable dependencies on this rule when the Vulcan SDK isn't installed. Building or depending on it will only result in a (late) error.

I'd like to be able to build as much as possible without the SDK installed. Not a big deal if the rule is gone or manual though, as long as when the SDK *is* installed, it comes back automatically?

llvm-bazel/WORKSPACE
GMNGeoffrey: Ah you want to condition the headers on the SDK being installed? The headers are actually required for building, so that would make building this require the SDK, which doesn't seem right to me. Bazel is lazy on this: it only runs the repo rule if you try to build a thing that depends on it.

llvm-bazel/llvm-project-overlay/mlir/BUILD
GMNGeoffrey: If we disable it when you don't have the SDK, then you can't build it even if you want to test it building and don't have the SDK installed. As it stands, you don't need the SDK installed to build it.

llvm-bazel/WORKSPACE
chandlerc: I understand that it'll be lazy, but if i build //... and don't have the SDK installed, it'll still pull headers.

Maybe I'm to nervous about pulling ever more dependencies, but I am a bit nervous. ;]

Will it really *always* build without the SDK installed? Even link? Even in dynamic mode? I find this somewhat surprising....

GMNGeoffrey: I think so? But I can't really guarantee. I'm mostly going off of the comments from people before me who actually understand what Vulkan is doing here :-D

I think a policy of "every target that transitively depends on an external dependency must be marked manual" would deal with your concern though, right? That's a bit annoying to enforce, although it's only like 5 targets right now, so probably nbd. Maybe bazel platform restrictions would help here, though I haven't played with them much.

My goal is that you should be able to at least *try* to build and query this target without the vulkan_sdk installed, which manual accomplishes. One of the things I don't like about Bazel selects is that they don't even let you try to do something wild. Like maybe this doesn't build for android, but I should be able to at least tell it to build for android explicitly and see what errors I get.

chandlerc: There are really two issues here:
1. How do we disable unsupported rules?
2. What criteria should we use for disabling Vulkan rules?

For the first, I agree with you that tagging as manual is nicer (if it works) so that you *can* see what errors result even when the config isn't set up in the expected way.

For the second, I think it'd be good to be a bit more conservative in enabling rules with these kinds of dependencies. Specifically, if there *is* some system aspect necessary, I'd only enable them when that system aspect is present (and only then pull the archive of headers, etc.)

GMNGeoffrey: I think for now tagging these as manual gives us something to work with that will enable integration with IREE so we can start dogfooding these there without making these less usable. We can continue to discuss the best way to handle this and other dependencies. SG?

README.md
chandlerc: Minor point, but I'd give some examples here... just feel like it'll be easier for users than finding the demos. But not a big deal, also would be fine to follow-up with docs later.

llvm-bazel/WORKSPACE
chandlerc: Sounds good.

llvm-bazel/llvm-project-overlay/mlir/BUILD
chandlerc: I'm fine with this for now, but I'd still like a better way for consumers of the Bazel rules to avoid accidentally ending up downloading and building against the headers.

As one user who is especially concerned with things like licenses of dependencies (much like LLVM itself is), I would really like a way to ensure that *no* dependencies other than llvm-bazel and the llvm-project are fetched in my Bazel build. Maybe that's a weird desire, but I'd like to see if there is a reasonably simple way to support that purely for simplicity of humans reasoning about these things. =] Anyways, fine to explore this in follow-ups, especially given the laziness.

chandlerc (approved):
LGTM

llvm-bazel/llvm-project-overlay/mlir/BUILD
GMNGeoffrey: On a project level, this is super easy. Just don't configure a repo rule for these dependencies. The WORKSPACE file for llvm-bazel isn't used by dependent projects. That seems much more robust than whether the SDK happens to be installed in terms of making sure you don't accidentally mix licenses.

README.md
GMNGeoffrey: You mean a usage demo? I can do that, but a demo that builds is always preferable to a demo in documentation that gets outdated IMO. Seems orthogonal to this PR regardless so going to defer that.

PR #24, opened 2020-09-21
Remove warning configuration from the BUILD file and enhance the

chandlerc:
`.bazelrc` to allow more easy build and test with both Clang and GCC.

I've also added some documentation to avoid simple mistakes, and
enhanced the way `NDEBUG` is defined to be explicit using the `defines`
attribute. While all the C++ toolchains I'm aware of in Bazel already
define this macro in exactly this way, it seems useful to make the build
self documenting by doing it explicitly here. This doesn't preclude
using `--copt=-UNDEBUG` to override it on the command line where needed.

The cleanups to the `.bazelrc` are somewhat opinionated on my part. ;]
It seems really good to make it easy to test different build
configurations here and so getting something a bit more vanilla (rather
than forcing `-c opt` by default seems good. If we want optimizations
enabled by default, a better way to do that is a custom toolchain that
sets up fastbuild with the desired flags, but for this project it
doesn't seem worth it.

### DIFF
--- a/llvm-bazel/.bazelrc
+++ b/llvm-bazel/.bazelrc
@@ -8,21 +8,48 @@
 ###############################################################################
 # Prevent invalid caching if input files are modified during a build.
 build --experimental_guard_against_concurrent_changes
-# Default to optimized builds
-# Override via: "-c dbg" or --compilation_mode=dbg
-build --compilation_mode=opt
 
 ###############################################################################
 # Options for "generic_clang" builds: these options should generally apply to
-# either clang or gcc and are curated based on need.
+# builds using a Clang-based compiler, and default to the `clang` executable on
+# the `PATH`. While these are provided for convenience and may serve as a
+# reference, it would be preferable for users to configure an explicit C++
+# toolchain instead of relying on `.bazelrc` files.
 ###############################################################################
+
+# Set the default compiler to the `clang` binary on the `PATH`.
+build:generic_clang --repo_env=CC=clang
 
 # C++14 standard version is required.
 build:generic_clang --cxxopt=-std=c++14 --host_cxxopt=-std=c++14
 
-# Default to adding back asserts in optimized builds.
-# This is a good compromise between runtime and debugability.
-build:generic_clang --copt=-UNDEBUG
+# Use `-Wall` and `-Werror` for Clang.
+build:generic_clang --copt=-Wall --copt=-Werror --host_copt=-Wall --host_copt=-Werror
+
+###############################################################################
+# Options for "generic_gcc" builds: these options should generally apply to
+# builds using a GCC-based compiler, and default to the `gcc` executable on
+# the `PATH`. While these are provided for convenience and may serve as a
+# reference, it would be preferable for users to configure an explicit C++
+# toolchain instead of relying on `.bazelrc` files.
+###############################################################################
+
+# Set the default compiler to the `gcc` binary on the `PATH`.
+build:generic_gcc --repo_env=CC=gcc
+
+# C++14 standard version is required.
+build:generic_gcc --cxxopt=-std=c++14 --host_cxxopt=-std=c++14
+
+# Disable GCC warnings that are noisy and/or false positives on LLVM code.
+# These need to be global as some code triggering these is in header files.
+build:generic_gcc --copt=-Wno-unused-parameter --host_copt=-Wno-unused-parameter
+build:generic_gcc --copt=-Wno-comment --host_copt=-Wno-comment
+build:generic_gcc --cxxopt=-Wno-class-memaccess --host_cxxopt=-Wno-class-memaccess
+build:generic_gcc --copt=-Wno-maybe-uninitialized --host_copt=-Wno-maybe-uninitialized
+build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading-indentation
+
+# Use `-Werror` for GCC to make sure warnings don't slip past.
+build:generic_gcc --copt=-Werror --host_copt=-Werror
 
 # The user.bazelrc file is not checked in but available for local mods.
 # Always keep this at the end of the file so that user flags override.
--- a/llvm-bazel/llvm-project-overlay/llvm/BUILD
+++ b/llvm-bazel/llvm-project-overlay/llvm/BUILD
@@ -14,20 +14,13 @@ exports_files(["LICENSE.TXT"])
 
 llvm_host_triple = "x86_64-unknown-linux_gnu"
 
-# FIXME: Would be nice to have a select here on the compiler so that we can use
-# a more minimal list for Clang than we need for GCC. Currently this merges the
-# two lists. However, that results in warning flags that aren't understood by
-# the compiler. For GCC, this will only result in extra warning output when
-# a warning or error occurs for some other reason. For Clang, we disable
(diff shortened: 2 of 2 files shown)


llvm-bazel/.bazelrc
```diff
@@ -8,21 +8,48 @@
 ###############################################################################
 # Prevent invalid caching if input files are modified during a build.
 build --experimental_guard_against_concurrent_changes
-# Default to optimized builds
-# Override via: "-c dbg" or --compilation_mode=dbg
-build --compilation_mode=opt
```
GMNGeoffrey, line 13: Yeah this was taken from IREE and I think was based on Stella's opinionated bazelrc :-D I believe the logic there was that the time savings from fastbuild are usually outweighed by the extra test time, but given that we don't actually have any tests in this repo yet, it doesn't make much sense here :grin:

llvm-bazel/.bazelrc
```diff
@@ -8,21 +8,48 @@
...
 
 ###############################################################################
 # Options for "generic_clang" builds: these options should generally apply to
-# either clang or gcc and are curated based on need.
+# builds using a Clang-based compiler, and default to the `clang` executable on
+# the `PATH`. While these are provided for convenience and may serve as a
+# reference, it would be preferable for users to configure an explicit C++
+# toolchain instead of relying on `.bazelrc` files.
```
GMNGeoffrey, line 17: A bit confused about who "users" is here. If I'm a person who wants to just build this repo are you saying I should be setting up a toolchain? Or by users do you mean dependent projects?

llvm-bazel/.bazelrc
```diff
@@ -8,21 +8,48 @@
...
 ###############################################################################
 
+# Set the default compiler to the `clang` binary on the `PATH`.
+build:generic_clang --repo_env=CC=clang
```
GMNGeoffrey, line 21: `CXX=clang++` too?

Is this necessary at all though? Repo rules already see the whole environment, so setting `CC=clang` in your environment should already cover this. I guess this is more explicit and still overridable with a user bazelrc or explicit command line pointing to another clang binary.

llvm-bazel/.bazelrc
```diff
@@ -8,21 +8,48 @@
...
+# Disable GCC warnings that are noisy and/or false positives on LLVM code.
+# These need to be global as some code triggering these is in header files.
+build:generic_gcc --copt=-Wno-unused-parameter --host_copt=-Wno-unused-parameter
+build:generic_gcc --copt=-Wno-comment --host_copt=-Wno-comment
+build:generic_gcc --cxxopt=-Wno-class-memaccess --host_cxxopt=-Wno-class-memaccess
+build:generic_gcc --copt=-Wno-maybe-uninitialized --host_copt=-Wno-maybe-uninitialized
+build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading-indentation
+
+# Use `-Werror` for GCC to make sure warnings don't slip past.
+build:generic_clang --copt=-Werror --host_copt=-Werror
```
GMNGeoffrey: `-Wall` as well? I think if you put it above the explicit disable flags it will be all but those.

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -14,22 +14,29 @@ exports_files(["LICENSE.TXT"])
...
-    "-Wno-class-memaccess",
-    "-Wno-maybe-uninitialized",
-    "-Wno-misleading-indentation",
-    "-Wno-unknown-warning-option",
 ]
 
+# Allow configuring attributes based on build mode.
+config_setting(
+    name = "dbg_build",
+    values = {"compilation_mode": "dbg"},
```
GMNGeoffrey: Huh surprised these aren't defined in a library already...

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -583,6 +590,10 @@ cc_library(
         "include/llvm/Config/llvm-config.h",
     ],
     copts = llvm_copts,
+    defines = select({
+        ":opt_build": ["NDEBUG"],
```
GMNGeoffrey: Pretty sure this is what `--compilation_mode=opt` *means* (in addition `-O2`):

https://docs.bazel.build/versions/master/user-manual.html#flag--compilation_mode

llvm-bazel/.bazelrc
chandlerc: Users here I think means dependent projects -- my feeling is that directly building here is mostly for developers of these Bazel files. Happy to clarify if you see a good way?

llvm-bazel/.bazelrc
chandlerc: No need to set both `CXX` and `CC` from what I can see using Bazel, but let me know if I've missed something.

As for why setting it this way -- it makes it much easier for me to test that both Clang and GCC are working cleanly to handle it on the commandline as a config rather than relying on a particular environment.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: Maybe they are? I asked around and none was suggested.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: Good point... I was hesitating about relying on this, but I think you're right that we really should just assume this and not worry. If we want, can always add a test to verify.

llvm-bazel/.bazelrc
chandlerc: (Note, there was a bug here and it said `generic_clang` but this is for `generic_gcc`.)

I'm not super motivated to clean up GCC's `-Wall`, I just want to make sure that on-by-default warnings don't fire. If someone else needs this, we can do so, but I suspect more disabling will be necessary.

chandlerc:
Replied and updated, thanks!

llvm-bazel/.bazelrc
GMNGeoffrey: I guess if it works that's proof enough for me :-D

Makes sense. Maybe we should adopt this in IREE

GMNGeoffrey: The disadvantage here is that someone who has clang at /my/weird/path and has their CC set to that is going to have to set it again with an explicit repo_env call, which is maybe too bad. Probably not a big deal though.

GMNGeoffrey (approved):

llvm-bazel/.bazelrc
chandlerc: Yeah, but if they are doing that they can just pass no `--config` at all and it should work.

Lemme know if you'd like me to factor the configs into one that sets `CC` and another that adds flags so that people can set `CC` themselves if they want and then still pick up our suggested flags. As I suspect these are mostly for the use of those developing the llvm-bazel project, it didn't seem worth doing yet, but I'm happy to add it.

chandlerc:
Thanks, merging. Let me know if the remaining comment threads need anything. =]

llvm-bazel/.bazelrc
GMNGeoffrey: Nah this is fine for now

PR #25, opened 2020-09-27
Introduce a configurable dependency on zlib.

chandlerc:
This allows users to either explicitly choose how they depend on zlib or
to use an environment variable to select between dependencies. It
includes the mechanical components needed to support external
repositories in a hermetic/Bazel style, system library dependencies, or
disabling zlib entirely.

The local workspace is set up to use the environment variable based
configuration, and some convenient `--config` wrappers are added to
enable us to test things easily.

### DIFF
--- a/llvm-bazel/llvm-project-overlay/llvm/BUILD
+++ b/llvm-bazel/llvm-project-overlay/llvm/BUILD
@@ -649,6 +649,10 @@ cc_library(
     deps = [
         ":Demangle",
         ":config",
+        # We unconditionally depend on the custom LLVM zlib wrapper. This will
+        # be an empty library unless zlib is enabled, in which case it will
+        # both provide the necessary dependencies and configuration defines.
+        "@llvm_zlib//:zlib",
     ],
 )
 
--- a/llvm-bazel/zlib.bzl
+++ b/llvm-bazel/zlib.bzl
@@ -1,0 +1,112 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""Repository rules to configure the zlib used by LLVM.
+
+Most users should pick one of the explicit rules to configure their use of zlib
+with LLVM:
+- `llvm_zlib_external` will link against an external Bazel zlib repository.
+- `llvm_zlib_system` will link against the system zlib (non-hermetically).
+- 'llvm_zlib_disable` will disable zlib completely.
+
+If you would like to make your build configurable, you can use
+`llvm_zlib_from_env`. By default, this will disable zlib, but will inspect
+the environment variable (most easily set with a `--repo_env` flag to the
+Bazel invocation) `BAZEL_LLVM_ZLIB_STRATEGY`. If it is set to `external`,
+then it will behave the same as `llvm_zlib_external`. If it is set to
+`system` then it will behave the same as `llvm_zlib_system`. Any other
+setting will disable zlib the same as not setting it at all.
+"""
+
+def _llvm_zlib_external_impl(repository_ctx):
+    repository_ctx.template(
+        "BUILD",
+        repository_ctx.attr._external_build_template,
+        substitutions = {
+            "@external_zlib_repo//:zlib_rule": str(repository_ctx.attr.external_zlib),
+        },
+        executable = False,
+    )
+
+llvm_zlib_external = repository_rule(
+    implementation = _llvm_zlib_external_impl,
+    attrs = {
+        "_external_build_template": attr.label(
+            default = Label("//deps_impl:zlib_external.BUILD"),
+            allow_single_file = True,
+        ),
+        "external_zlib": attr.label(
+            doc = "The dependency that should be used for the external zlib library.",
+            mandatory = True,
+        ),
+    },
+)
+
+def _llvm_zlib_system_impl(repository_ctx):
+    repository_ctx.template(
+        "BUILD",
+        repository_ctx.attr._system_build_template,
+        executable = False,
+    )
+
+# While it may seem like this needs to be local, it doesn't actually inspect
+# any local state, it just configures to build against that local state.
+llvm_zlib_system = repository_rule(
+    implementation = _llvm_zlib_system_impl,
+    attrs = {
+        "_system_build_template": attr.label(
+            default = Label("//deps_impl:zlib_system.BUILD"),
+            allow_single_file = True,
+        ),
+    },
+)
+
+def _llvm_zlib_disable_impl(repository_ctx):
+    repository_ctx.template(
+        "BUILD",
+        repository_ctx.attr._disable_build_template,
+        executable = False,
+    )
+
+llvm_zlib_disable = repository_rule(
+    implementation = _llvm_zlib_disable_impl,
+    attrs = {
+        "_disable_build_template": attr.label(
+            default = Label("//deps_impl:zlib_disable.BUILD"),
+            allow_single_file = True,
+        ),
+    },
+)
+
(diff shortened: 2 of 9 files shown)


llvm-bazel/zlib_disable.BUILD
```diff
@@ -0,0 +1,10 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
```
GMNGeoffrey: I think these should go under third_party_build to tuck them away a bit more. Alternatively, there's enough files here that we should maybe have a zlib directory.

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -649,6 +649,10 @@ cc_library(
     deps = [
         ":Demangle",
         ":config",
+        # We unconditionally depend on the custom LLVM zlib wrapper. This will
+        # be an empty library unless zlib is enabled, in which case it will
+        # both provide the necessary dependencies and configuration defines.
+        "@llvm_zlib//:zlib",
```
GMNGeoffrey, line 655: Shouldn't this be part of config instead? That's where the zlib defines are usually, I think?

llvm-bazel/zlib.bzl
```diff
@@ -0,0 +1,113 @@
...
+        repository_ctx.attr._external_build_template,
+        substitutions = {
+            "@external_zlib_repo//:zlib_rule": str(repository_ctx.attr.external_zlib),
+        },
+        executable = False,
+    )
+
+llvm_zlib_external = repository_rule(
+    implementation = _llvm_zlib_external_impl,
+    local = True,
```
GMNGeoffrey: This only uses declared dependencies, so doesn't need to be declared local. https://docs.bazel.build/versions/master/skylark/repository_rules.html#when-is-the-implementation-function-executed

llvm-bazel/zlib.bzl
```diff
@@ -0,0 +1,113 @@
...
+def _llvm_zlib_system_impl(repository_ctx):
+    repository_ctx.template(
+        "BUILD",
+        repository_ctx.attr._system_build_template,
+        executable = False,
+    )
+
+llvm_zlib_system = repository_rule(
+    implementation = _llvm_zlib_system_impl,
+    local = True,
```
GMNGeoffrey: Same here on being local, I think. Despite the fact that the configured rule is non-hermetic in that it uses the system zlib, nothing about the configuration needs to be rerun.

llvm-bazel/zlib_system.BUILD
```diff
@@ -0,0 +1,13 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Wrapper library for the system's zlib. Using this only works if the toolchain
+# already has the relevant header search and library search paths configured.
+# It also sets the relevant LLVM `#define`s to enoble zlib.
```
GMNGeoffrey: Typo

```suggestion
# It also sets the relevant LLVM `#define`s to enable zlib.
```

llvm-bazel/zlib_disable.BUILD
```diff
@@ -0,0 +1,10 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Empty stub library. This doesn't include zlib, but also doesn't set the LLVM
```
GMNGeoffrey: Nit: "but also" seems like strange wording here

```suggestion
# Empty stub library. This doesn't include zlib and doesn't set the LLVM
```

llvm-bazel/zlib.bzl
```diff
@@ -0,0 +1,113 @@
...
+def _llvm_zlib_disable_impl(repository_ctx):
+    repository_ctx.template(
+        "BUILD",
+        repository_ctx.attr._disable_build_template,
+        executable = False,
+    )
+
+llvm_zlib_disable = repository_rule(
+    implementation = _llvm_zlib_disable_impl,
+    local = True,
```
GMNGeoffrey: Also not local

llvm-bazel/zlib.bzl
```diff
@@ -0,0 +1,113 @@
...
+    if zlib_strategy == "external":
+        _llvm_zlib_external_impl(repository_ctx)
+    elif zlib_strategy == "system":
+        _llvm_zlib_system_impl(repository_ctx)
+    else:
+        _llvm_zlib_disable_impl(repository_ctx)
+
+llvm_zlib_from_env = repository_rule(
+    implementation = _llvm_zlib_from_env_impl,
+    local = True,
```
GMNGeoffrey: Declare `BAZEL_LLVM_ZLIB_STRATEGY` in `environ` and then I think this can be not local as well.

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: I sunk it as low as I could in the dependency graph... None of this goes into the public `llvm-config.h` so it shouldn't be surfaced in an ABI visible way, so I just tucked it here.

One reason to have this low in the dependency graph is that consumers of `config` that don't need zlib shouldn't get an artificial dependency on it.

llvm-bazel/zlib.bzl
chandlerc: At true, fixed!

llvm-bazel/zlib.bzl
chandlerc: oooo, didn't know about declaring environ, thanks!

llvm-bazel/zlib_disable.BUILD
chandlerc: I thought about `thrid_party_build` but that seems wrong: this doesn't actually build any third-party code, it is a wrapper only.

I'm happy to create a directory if you want, but didn't seem quite worth it (4 files) to me.

llvm-bazel/zlib.bzl
chandlerc: Interesting. I'll leave a comment to explain why. =]

chandlerc (commented):
Thanks for the review, should be updated!

llvm-bazel/zlib_disable.BUILD
GMNGeoffrey: I named the directory that for "third-party build files" :grin: The first thing it contained was a build file for zlib. We can totally call that directory something else. `dependency_configuration`? But with zlib and terminfo up here it seems like a lot of files

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Ok. Wasn't sure if we needed the defines higher up, since they're usually defined in config.h. If that was the case, we could have separate zlib and zlib_defines rules.

GMNGeoffrey (approved):

llvm-bazel/zlib_disable.BUILD
GMNGeoffrey: Some discussion in discord. We're going to put "real" build files that actually compile some third-party source code (and therefore should carry a license header) in `third_party_build` and put implementation details of dependency configuration (like build files with empty rules) in `deps_impl`. We'll leave the bzl configurations in `llvm-bazel` as they are actually "interesting"

chandlerc: As discussed on chat, we have `third_party_build` for rules that actually build third-party code (and thus maybe need a license).

These bits are just implementation details and don't need that. Moved them to a `deps_impl` directory to make it clear.

GMNGeoffrey (approved):

PR #26, opened 2020-09-28
Add support for linking against terminfo.

chandlerc:
I'm not aware of an effective external Bazel-style dependency, so this
only adds support for detecting and configuring a system-library. This
is somewhat more complex than other libraries because there are a few
different libraries that provide these APIs.

I've tried to ensure sufficient configuration hooks for these rules to
allow different users of LLVM to enable or not this functionality.

### DIFF
--- a/llvm-bazel/terminfo.bzl
+++ b/llvm-bazel/terminfo.bzl
@@ -1,0 +1,195 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""Repository rules to configure the terminfo used by LLVM.
+
+Most users should pick one of the explicit rules to configure their use of terminfo
+with LLVM:
+- `llvm_terminfo_system` will detect and link against a terminfo-implementing
+  system library (non-hermetically).
+- 'llvm_terminfo_disable` will disable terminfo completely.
+
+If you would like to make your build configurable, you can use
+`llvm_terminfo_from_env`. By default, this will disable terminfo, but will
+inspect the environment variable (most easily set with a `--repo_env` flag to
+the Bazel invocation) `BAZEL_LLVM_TERMINFO_STRATEGY`. If it is set to
+`system` then it will behave the same as `llvm_terminfo_system`. Any other
+setting will disable terminfo the same as not setting it at all.
+"""
+
+def _llvm_terminfo_disable_impl(repository_ctx):
+    repository_ctx.template(
+        "BUILD",
+        repository_ctx.attr._disable_build_template,
+        executable = False,
+    )
+
+_terminfo_disable_attrs = {
+    "_disable_build_template": attr.label(
+        default = Label("//deps_impl:terminfo_disable.BUILD"),
+        allow_single_file = True,
+    ),
+}
+
+llvm_terminfo_disable = repository_rule(
+    implementation = _llvm_terminfo_disable_impl,
+    attrs = _terminfo_disable_attrs,
+)
+
+def _find_c_compiler(repository_ctx):
+    """Returns the path to a plausible C compiler.
+
+    This routine will only reliably work on roughly POSIX-y systems as it
+    ultimately falls back on the `cc` binary. Fortunately, the thing we are
+    trying to use it for (detecting if a trivial source file can compile and
+    link against a particular library) requires very little.
+    """
+    cc = repository_ctx.which(repository_ctx.os.environ.get("CC"))
+    if not cc:
+        # Look for Clang, GCC, and the POSIX / UNIX specified C compiler
+        # binaries.
+        for compiler in ["clang", "gcc", "c99", "c89", "cc"]:
+            cc = repository_ctx.which(compiler)
+            if cc:
+                break
+    return cc
+
+def _try_link(repository_ctx, cc, source, linker_flags):
+    """Returns `True` if able to link the source with the linker flag.
+
+    Given a source file that contains references to library routines, this
+    will check that when linked with the provided linker flag, those
+    references are successfully resolved. This routine assumes a generally
+    POSIX-y and GCC-ish compiler and environment and shouldn't be expected to
+    work outside of that.
+    """
+    cmd = [
+        cc,
+        # Force discard the linked executable.
+        "-o",
+        "/dev/null",
+        # Leave language detection to the compiler.
+        source,
+    ]
+
+    # The linker flag must be valid for a compiler invocation of the link step,
+    # so just append them to the command.
+    cmd += linker_flags
+    exec_result = repository_ctx.execute(cmd, timeout = 20)
+    return exec_result.return_code == 0
+
+def _llvm_terminfo_system_impl(repository_ctx):
+    # LLVM doesn't need terminfo support on Windows, so just disable it.
+    if repository_ctx.os.name.lower().find("windows") != -1:
+        _llvm_terminfo_disable_impl(repository_ctx)
+        return
+
+    if len(repository_ctx.attr.system_linkopts) > 0:
+        linkopts = repository_ctx.attr.system_linkopts
+    else:
+        required = repository_ctx.attr.system_required
+
+        # Find a C compiler we can use to detect viable linkopts on this system.
+        cc = _find_c_compiler(repository_ctx)
+        if not cc:
+            if required:
+                fail("Failed  to find a C compiler executable")
+            else:
+                _llvm_terminfo_disable_impl(repository_ctx)
+                return
+
+        # Get the source file we use to detect successful linking of terminfo.
+        source = repository_ctx.path(repository_ctx.attr._terminfo_test_source)
+
+        # Collect the candidate linkopts and wrap them into a list. Ideally,
+        # these would be provided as lists, but Bazel doesn't currently
+        # support that. See: https://github.com/bazelbuild/bazel/issues/12178
+        linkopts_candidates = [[x] for x in repository_ctx.attr.candidate_system_linkopts]
+
+        # For each candidate, try to use it to link our test source file.
+        for linkopts_candidate in linkopts_candidates:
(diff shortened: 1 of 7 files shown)


chandlerc:
Note that this is intended as a follow-up to #25 -- Let me know if you want me to create an branch in the upstream repository to allow the diff to be reviewed w/o that.

GMNGeoffrey:
> Note that this is intended as a follow-up to #25 -- Let me know if you want me to create an branch in the upstream repository to allow the diff to be reviewed w/o that.

I can view the diff per-commit :-)

llvm-bazel/terminfo_system.BUILD
```diff
@@ -0,0 +1,15 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
```
GMNGeoffrey: I think it's customary to name template files `.tmpl`

llvm-bazel/terminfo_disable.BUILD
```diff
@@ -0,0 +1,10 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Empty stub library. This doesn't include any terminfo library, but also
+# doesn't set the LLVM `#define`s to enable usage of terminfo.
```
GMNGeoffrey:
```suggestion
# Empty stub library. This doesn't include any terminfo library and doesn't set
# the LLVM `#define`s to enable usage of terminfo.
```

llvm-bazel/terminfo.bzl
```diff
@@ -0,0 +1,217 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""Repository rules to configure the terminfo used by LLVM.
+
+Most users should pick one of the explicit rules to configure their use of terminfo
+with LLVM:
+- `llvm_terminfo_external` will link against an external Bazel terminfo repository.
```
GMNGeoffrey: Can you update the docs here to indicate that external really isn't supported?

llvm-bazel/terminfo.bzl
```diff
@@ -0,0 +1,217 @@
...
+}
+
+llvm_terminfo_disable = repository_rule(
+    implementation = _llvm_terminfo_disable_impl,
+    local = True,
+    attrs = _terminfo_disable_attrs,
+)
+
+def _llvm_terminfo_from_env_impl(repository_ctx):
+    terminfo_strategy = repository_ctx.os.environ.get("BAZEL_LLVM_TERMINFO_STRATEGY")
```
GMNGeoffrey, line 183: I think you could add `BAZEL_LLVM_TERMINFO_STRATEGY` and `CC` to [`environ`](https://docs.bazel.build/versions/master/skylark/lib/globals.html#repository_rule.environ), mostly just as documentation

llvm-bazel/terminfo.bzl
```diff
@@ -0,0 +1,217 @@
...
+    if terminfo_strategy == "external":
+        pass #_llvm_terminfo_external_impl(repository_ctx)
```
GMNGeoffrey: Error?

llvm-bazel/terminfo.bzl
```diff
@@ -0,0 +1,217 @@
...
+            else:
+                _llvm_terminfo_disable_impl(repository_ctx)
+                return
+
+        # Get the source file we use to detect successful linking of terminfo.
+        source = repository_ctx.path(repository_ctx.attr._terminfo_test_source)
+
+        # Collect the candidate linkopts and wrap them into a list. Ideally,
+        # these would be provided as lists, but Bazel doesn't currently
+        # support that. See feature request #12178.
```
GMNGeoffrey: Can you link to the Bazel feature request. Otherwise this implies an issue in this repository.

```suggestion
        # support that. See https://github.com/bazelbuild/bazel/issues/12178
```

GMNGeoffrey: Why should they be lists? Would a user need to potentially pass multiple link options as a single candidate?

llvm-bazel/terminfo_system.BUILD
```diff
@@ -0,0 +1,15 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Wrapper library for some system terminfo. Using this only works if the
+# toolchain already has the relevant library search paths configured. It also
+# sets the relevant LLVM `#define`s to enoble using terminfo.
+cc_library(
+    name = "terminfo",
+    defines = ["LLVM_ENABLE_TERMINFO=1"],
+    # Note that we will replace the exact library here with the name detected
+    # on the system.
```
GMNGeoffrey: We will? I think I maybe don't understand this sentence. We're replacing the linkopts with the ones detected on the system and I guess that's basically the name of the terminfo, but we're not replacing the library...

llvm-bazel/terminfo_test.c
```diff
@@ -0,0 +1,17 @@
...
+extern struct term *set_curterm(struct term *termp);
+extern int del_curterm(struct term *termp);
+extern int tigetnum(char *capname);
+
+int main() {
+  setupterm(0, 0, 0);
+  set_curterm(0);
+  del_curterm(0);
+  tigetnum(0);
+}
```
GMNGeoffrey: Nit. Missing trailing newline

llvm-bazel/terminfo.bzl
```diff
@@ -0,0 +1,217 @@
...
+    link against a particular library) requires very little.
+    """
+    cc = repository_ctx.which(repository_ctx.os.environ.get("CC"))
+    if not cc:
+        cc = repository_ctx.which("clang")
+    if not cc:
+        cc = repository_ctx.which("gcc")
+    if not cc:
+        cc = repository_ctx.which("cc")
+    return cc
```
GMNGeoffrey: I feel like it would be pretty reasonable to just require that CC be set and error otherwise. Less magic and something that needs to be set anyway.

llvm-bazel/terminfo.bzl
chandlerc: Oh, sorry, I meant to remove all of this when I realized there wasn't an easy BSD package we could use to support this.

In fact, I had updated all of this and failed to save the file! Oops!

llvm-bazel/terminfo.bzl
chandlerc: Unfortunately, `CC` is not typically set. =[ The fact that this project uses it is somewhat specific. Other projects use a custom C++ toolchain configuration rather than any `CC` environment variable, and on typical Linux systems `CC` is not set. So unless Bazel itself *introduces* the environment variable for us to depend on in this way, I think we need some fallback.

The SUS/POSIX "way" is `c99`, `c89`, and `cc` binaries in the `PATH`. =/ I should add `c99` and `c89`. I'm pretty confident that this kind of fallback list is roughly what you want in terms of maximum portability across vaguely Unix-like systems (spanning Linux to macOS).

llvm-bazel/terminfo.bzl
chandlerc: Potentially.

As an example `["-L/path/to/search", "-llibrary_name"]`.

llvm-bazel/terminfo.bzl
chandlerc: Nice!

Maybe this is even enough to drop `local` and instead marke this as `configure`?

llvm-bazel/terminfo.bzl
chandlerc: Sorry, removing this too didn't get saved!

llvm-bazel/terminfo_system.BUILD
chandlerc: All these files are technically templates.

However, if I name them as `.tpl` or `.tmpl` I lose all syntax highlighting. Seemed nicer to name them this way and get the highlighting?

llvm-bazel/terminfo_system.BUILD
chandlerc: Sorry, yes, stale comment. Updated.

llvm-bazel/terminfo_test.c
chandlerc: Done.

chandlerc (commented):
Thanks, PTAL!

llvm-bazel/terminfo_system.BUILD
GMNGeoffrey: Fair enough. I think frequently they're templates with no substitutions though, which seemed not worth the extension. Another option would be `.tmpl.BUILD`

llvm-bazel/terminfo.bzl
GMNGeoffrey: Yeah I thought I understood `configure` and `local` and then I read the docs again and decided maybe I didn't... I'm fine with whatever you think is right here and then we can come back and re-annotate if we learn what's actually right...

GMNGeoffrey (approved):

llvm-bazel/terminfo_system.BUILD
chandlerc: Let me know if you want me to make the change to `.tmpl.BUILD` or `.tpl.BUILD` (the latter abbreviation matches what I saw in the Bazel C++ toolchain, but I'm happy to go with whichever you prefer). I'll do that in a follow-up if alright though, as I can update the zlib files as well.

llvm-bazel/terminfo.bzl
chandlerc: I'm going with both set to true. The `configure` needs to be true to reflect the system, and the `local` I think needs to be true because the system is accessed outside of the specified inputs to the rule. But I'm mostly making this up. Can remove one if/when we hear back from Bazel folks about the correct settings here.

PR #29, opened 2020-10-01
Drop usage of marketplace push action

GMNGeoffrey:
This is failing for some mysterious reason at the moment. Basically all
it does is configure auth and call push, but the checkout action should
have already configured auth for us, so not much added value.


chandlerc (approved):

GMNGeoffrey:
This is going to fail on the push step right now because GitHub gets upset when you push too many refs :-/ https://github.com/isaacs/github/issues/1253. Tested with https://github.com/GMNGeoffrey/llvm-bazel/runs/1191014351. Still want to get this in though, and I'll manually update us to be closer to LLVM HEAD

PR #34, opened 2020-10-02
Make autointegrate action idempotent

GMNGeoffrey:
This allows us to trigger it on other events (e.g. a cron) to update
for new LLVM commits without constantly clobbering the auto-integrate
branch.

Also switches to a personal access token (for a bot) so that there
aren't issues pushing workflow files (which I ran into while testing)
and so that the action can trigger webhooks for other automation (e.g.
builds).

I'm experiencing some intermittent failures that look like
https://github.com/actions/checkout/issues/363, but those were
happening before this change as well.

Tested:
Ran on my fork:
https://github.com/GMNGeoffrey/llvm-bazel/runs/1201208655
Second run:
https://github.com/GMNGeoffrey/llvm-bazel/runs/1201226878

### DIFF
--- a/.github/workflows/auto_integrate.yml
+++ b/.github/workflows/auto_integrate.yml
@@ -42,6 +42,7 @@ jobs:
           else
             echo "${INTEGRATION_BRANCH?} already branches from ${MERGING_BRANCH?}"
           fi
+          git checkout "${INTEGRATION_BRANCH?}"
           git submodule update
       - name: Running auto integration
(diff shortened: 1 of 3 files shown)


chandlerc (approved):
Can't say I fully understand this, but seems plausible to me. =]

GMNGeoffrey:
> Can't say I fully understand this, but seems plausible to me. =]

If the HEAD commit of main is already in the auto-integrate branch, then don't clobber-merge it in again :grin:

PR #36, opened 2020-10-03
Configure RBE

GMNGeoffrey:
This allows remotely executing and caching actions.

The primary motivation is to be able to have a cluster of build agents
all sharing a cache.

We have one issue with using the default RBE toolchain: it doesn't have
xml2-dev, so we get link failures for WindowsManifest. For now, I've
marked the target manual. I also played around with the
`no-remote-exec` tag, but there are two issues: It doesn't actually
*work* without an experimental flag
https://github.com/bazelbuild/bazel/issues/7988. Even turning that on,
Bazel tries to look for the clang binary at the path it would be on the
remote on the local machine, which doesn't work so well.

> src/main/tools/linux-sandbox-pid1.cc:426:
"execvp(/usr/local/bin/clang, 0x1f81180)": No such file or directory

To get this working, we can use a custom docker container for our
toolchain that includes the right dependencies. This isn't too hard to
set up. But before I do it, do we actually care about this binary right
now? I can punt if we don't.

### DIFF
--- a/llvm-bazel/.bazelrc
+++ b/llvm-bazel/.bazelrc
@@ -62,6 +62,60 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
 # Use `-Werror` for GCC to make sure warnings don't slip past.
 build:generic_gcc --copt=-Werror --host_copt=-Werror
 
+###############################################################################
+# Configuration for building remotely using Remote Build Execution (RBE)
+# https://cloud.google.com/remote-build-execution/
+# Based on https://github.com/bazelbuild/bazel-toolchains/blob/master/bazelrc/bazel-1.0.0.bazelrc
+###############################################################################
+
+build:rbe --remote_instance_name=projects/llvm-bazel/instances/default_instance
+
+# Depending on how many machines are in the remote execution instance, setting
+# this higher can make builds faster by allowing more jobs to run in parallel.
+# Setting it too high can result in jobs that timeout, however, while waiting
+# for a remote machine to execute them.
+build:rbe --jobs=50
+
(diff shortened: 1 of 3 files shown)


chandlerc:
> This allows remotely executing and caching actions.
> 
> The primary motivation is to be able to have a cluster of build agents
> all sharing a cache.
> 
> We have one issue with using the default RBE toolchain: it doesn't have
> xml2-dev, so we get link failures for WindowsManifest. For now, I've
> marked the target manual.

Doh, we should really handle this the same way as zlib. Do you want to wrap it up that way? Want me to?

Clearly doesn't need to be in this patch...

> I also played around with the
> `no-remote-exec` tag, but there are two issues: It doesn't actually
> _work_ without an experimental flag
> [bazelbuild/bazel#7988](https://github.com/bazelbuild/bazel/issues/7988). Even turning that on,
> Bazel tries to look for the clang binary at the path it would be on the
> remote on the local machine, which doesn't work so well.
> 
> > src/main/tools/linux-sandbox-pid1.cc:426:
> > "execvp(/usr/local/bin/clang, 0x1f81180)": No such file or directory
> 
> To get this working, we can use a custom docker container for our
> toolchain that includes the right dependencies. This isn't too hard to
> set up. But before I do it, do we actually care about this binary right
> now? I can punt if we don't.

chandlerc (approved):

GMNGeoffrey:
> Doh, we should really handle this the same way as zlib. Do you want to wrap it up that way? Want me to?
> 
> Clearly doesn't need to be in this patch...

I can take a look in a follow-up. Is libxml optional? Or we can grab it from source I guess?

PR #37, opened 2020-10-03
Fix the terminfo detection to work without a `CC` env variable.

chandlerc:
I thought the `which` API would be well behaved here, but it produces
a hard error on a `None` argument. Using a default of `""` doesn't work
either because of https://github.com/bazelbuild/bazel/issues/12216

### DIFF
--- a/llvm-bazel/terminfo.bzl
+++ b/llvm-bazel/terminfo.bzl
@@ -45,7 +45,10 @@ def _find_c_compiler(repository_ctx):
     trying to use it for (detecting if a trivial source file can compile and
     link against a particular library) requires very little.
     """
-    cc = repository_ctx.which(repository_ctx.os.environ.get("CC"))
+    cc_env = repository_ctx.os.environ.get("CC")
+    cc = None
+    if cc_env:
+        cc = repository_ctx.which(cc_env)
     if not cc:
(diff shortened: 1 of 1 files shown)


llvm-bazel/terminfo.bzl
```diff
@@ -45,7 +45,10 @@ def _find_c_compiler(repository_ctx):
     trying to use it for (detecting if a trivial source file can compile and
     link against a particular library) requires very little.
     """
-    cc = repository_ctx.which(repository_ctx.os.environ.get("CC"))
+    cc_env = repository_ctx.os.environ.get("CC")
+    cc = None
+    if cc_env:
+        cc = repository_ctx.which(cc_env)
```
GMNGeoffrey, line 51: Just a note that this will also skip if `CC=""`. I *think* that's what we want, but something to bear in mind. The python style guide says to always make None comparisons explicit for that reason https://google.github.io/styleguide/pyguide.html#2144-decision

GMNGeoffrey (approved):

llvm-bazel/terminfo.bzl
chandlerc: Yeah, we actually want that for now due to the Bazel bug. And even when fixed, there is no point, do this still seems good to avoid a pointless call to `which`.

PR #41, opened 2020-10-06
Integrate LLVM at llvm/llvm-project@5e4409f30817

GMNGeoffrey:
[5e4409f30817](https://github.com/llvm/llvm-project/commit/5e4409f30817)


GMNGeoffrey:
This is a separate PR from https://github.com/google/llvm-bazel/pull/40 because I want to have working BUILD files for both those commits. With https://github.com/google/llvm-bazel/pull/42, we should be able to do this sort of thing in a single PR with multiple commits that we rebase and merge.

chandlerc (approved):

PR #48, opened 2020-10-10
Rework llvm configuration

GMNGeoffrey:
The idea of string-substituting the .cmake file was to keep things
simple and ensure that we detected changes to the .cmake file that we
would need to pick up.

As we need to move to selects based on platform this is going to get
super verbose. We've also already pulled some things out into Bazel
defines (e.g. zlib and terminfo features).

As for detecting drift, it only sort of worked, since it wouldn't
detect if a variable was removed, e.g.
https://github.com/llvm/llvm-project/commit/31e5f7120bdd. This leaves
us with some confusing tech debt for later.

Instead this change splits the jobs of setting the appropriate defines
between hardcoded `.h` files, for things that don't change based on
platform, and Bazel defines based on selects, for those that do. We
could eventually move everything into Bazel defines if we wanted to,
but this is maybe more work for things that don't change.

To kick things off, I extracted a few variables that I know will be
different on Windows, but I have no confidence that this is now
configured such that a Windows build would pass and I don't have a box
to test on.

Another approach would be to write a true equivalent to CMake's
`configure_file` and use that along with a dict for the defines.
@chandlerc expressed concerns about their being a python dependency in
the actual build stage, so this should probably be written in C[++].

For detecting drift we move from something that sort of does so to
something that definitely will: literal change-detector tests. To avoid
a loading time dependency on `bazel_skylib` in the actual project's
build rules (that we expect dependent projects to use), these tests are
in the root workspace.


llvm-bazel/llvm-project-overlay/llvm/config.bzl
```diff
@@ -0,0 +1,44 @@
...
+        [
+            "LLVM_HOST_TRIPLE=\\\"x86_64-pc-win32\\\"",
+            "LLVM_DEFAULT_TARGET_TRIPLE=\\\"x86_64-pc-win32\\\"",
+            "LLVM_NATIVE_ARCH=\\\"X86\\\"",
+        ] + win32_defines
+    ),
+    "//conditions:default": (
+        [
+            "LLVM_HOST_TRIPLE=\\\"x86_64-unknown-linux_gnu\\\"",
+            "LLVM_DEFAULT_TARGET_TRIPLE=\\\"x86_64-unknown-linux_gnu\\\"",
```
chandlerc:
```suggestion
            "LLVM_HOST_TRIPLE=\\\"x86_64-unknown-linux-gnu\\\"",
            "LLVM_DEFAULT_TARGET_TRIPLE=\\\"x86_64-unknown-linux-gnu\\\"",
```

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/config.h
```diff
@@ -0,0 +1,364 @@
+
+/*===------- llvm/Config/llvm-config.h - llvm configuration -------*- C -*-===*/
```
chandlerc:
```suggestion
/*===------- llvm/Config/llvm-config.h - llvm configuration -------*- C -*-===*/
```

(At least, I think? Or does the .cmake file actually have a blank line?)

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
```diff
@@ -0,0 +1,102 @@
...
+#define LLVM_USE_INTEL_JITEVENTS 0
+
+/* Define if we have the oprofile JIT-support library */
+#define LLVM_USE_OPROFILE 0
+
+/* Define if we have the perf JIT-support library */
+#define LLVM_USE_PERF 0
+
+/* Major version of the LLVM API */
+#define LLVM_VERSION_MAJOR 11
```
chandlerc:
```suggestion
#define LLVM_VERSION_MAJOR 12
```

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
```diff
@@ -0,0 +1,102 @@
...
+
+/* Minor version of the LLVM API */
+#define LLVM_VERSION_MINOR 0
+
+/* Patch version of the LLVM API */
+#define LLVM_VERSION_PATCH 0
+
+/* LLVM version string */
+#define LLVM_VERSION_STRING "11.0.0git"
```
chandlerc:
```suggestion
#define LLVM_VERSION_STRING "12.0.0git"
```

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
```diff
@@ -0,0 +1,102 @@
...
+   This is a C header that can be included in the llvm-c headers. */
+
+#ifndef LLVM_CONFIG_H
+#define LLVM_CONFIG_H
+
+/* Define if LLVM_ENABLE_DUMP is enabled */
+/* #undef LLVM_ENABLE_DUMP */
+
+/* Target triple LLVM will generate code for by default */
+#define LLVM_DEFAULT_TARGET_TRIPLE "x86_64-unknown-linux-gnu"
```
chandlerc: This should probably be based on selects to pave the way for non-x86...

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
```diff
@@ -0,0 +1,102 @@
...
+/* LLVM_ENABLE_THREADS defined in Bazel */
+
+/* Has gcc/MSVC atomic intrinsics */
+#define LLVM_HAS_ATOMICS 1
+
+/* Host triple LLVM will be executed on */
+#define LLVM_HOST_TRIPLE "x86_64-unknown-linux-gnu"
+
+/* LLVM architecture name for the native architecture, if available */
+#define LLVM_NATIVE_ARCH X86
```
chandlerc: These already are part of `select`s.

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
```diff
@@ -0,0 +1,102 @@
...
+#define LLVM_NATIVE_DISASSEMBLER LLVMInitializeX86Disassembler
+
+/* LLVM name for the native Target init function, if available */
+#define LLVM_NATIVE_TARGET LLVMInitializeX86Target
+
+/* LLVM name for the native TargetInfo init function, if available */
+#define LLVM_NATIVE_TARGETINFO LLVMInitializeX86TargetInfo
+
+/* LLVM name for the native target MC init function, if available */
+#define LLVM_NATIVE_TARGETMC LLVMInitializeX86TargetMC
```
chandlerc: To support non-x86 hosts, you'll want all of these to be in select land.

chandlerc (commented):
Awesome!

I have  a bunch of somewhat minor comments below.

One comment that may only be relevant in a follow-up -- given that the hand written file has everything in it from the .cmake file, I wonder if the change detector test should use some python (its just a test, so easier) and check that all the identifiers after a `#...` show up in the hand written file (and maybe even in the same order?)

Unsure if that would be substantially more work than the current change detector tests. The current set up seems fine too, and certainly fine for  now.

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/config.h
GMNGeoffrey: Nope it does not. It also doesn't have a license header (and probably this license header should say the right filename :-D)

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
GMNGeoffrey: Yeah it is actually part of the selects. I missed it here because it's *also* in config.h.cmake for some reason.

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
GMNGeoffrey: Yeah I was punting because everything was X86 for now, but moved these over.

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
GMNGeoffrey: Fixed

GMNGeoffrey (commented):
> One comment that may only be relevant in a follow-up -- given that the hand written file has everything in it from the .cmake file, I wonder if the change detector test should use some python (its just a test, so easier) and check that all the identifiers after a `#...` show up in the hand written file (and maybe even in the same order?)
> 
> Unsure if that would be substantially more work than the current change detector tests. The current set up seems fine too, and certainly fine for now.

Yeah that seems like it might be a good idea. Maybe in a follow-up The change-detector was intended as "this definitely covers all our bases". I think we could do something looser that might be less brittle. My thought was to wait and see in what ways the change-detector gets annoying though and modify it from there.

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/config.h
```diff
@@ -0,0 +1,363 @@
...
+/* LTDL_SHLIB_EXT defined in Bazel */
+
+/* Define to the address where bug reports for this package should be sent. */
+#define PACKAGE_BUGREPORT "https://bugs.llvm.org/"
+
+/* Define to the full name of this package. */
+#define PACKAGE_NAME "LLVM"
+
+/* Define to the full name and version of this package. */
+#define PACKAGE_STRING "LLVM 11.0.0git"
```
chandlerc:
```suggestion
#define PACKAGE_STRING PACKAGE_NAME " " LLVM_VERSION_STRING
```

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/config.h
```diff
@@ -0,0 +1,363 @@
...
+
+/* Define to the version of this package. */
+#define PACKAGE_VERSION "11.0.0git"
```
chandlerc:
```suggestion
#define PACKAGE_VERSION LLVM_VERSION_STRING
```

chandlerc (approved):
LGTM

PR #49, opened 2020-10-11
Explicitly tag targets that don't build on Build Kite

GMNGeoffrey:
Some targets are tagged manual for other reasons and manual really just
means "don't build this in wildcards", which is a human-focused
distinction. For build bots, we can now use `bazel query` (which
ignores "manual") piped to `bazel test` to build and test all targets
in parallel (more important as we get more tests).

Tested:
Locally ran
```shell
bazel query //... + @llvm-project//... \
  | xargs bazel test \
    --config=rbe \
    --test_output=errors \
    --test_tag_filters=-nobuildkite \
    --build_tag_filters=-nobuildkite
```

### DIFF
--- a/llvm-bazel/configure.bzl
+++ b/llvm-bazel/configure.bzl
@@ -10,4 +10,4 @@ load(":overlay_directories.bzl", "overlay_directories")
 OVERLAY_DIR = "llvm-project-overlay"
 
 def llvm_configure(name, overlay_path = OVERLAY_DIR, **kwargs):
-    overlay_directories(name=name, overlay_path=overlay_path, **kwargs)
+    overlay_directories(name = name, overlay_path = overlay_path, **kwargs)
--- a/llvm-bazel/llvm-project-overlay/mlir/BUILD
+++ b/llvm-bazel/llvm-project-overlay/mlir/BUILD
@@ -3289,7 +3289,7 @@ cc_binary(
     srcs = ["tools/mlir-vulkan-runner/vulkan-runtime-wrappers.cpp"],
     linkshared = True,
     tags = [
-        "manual",  # Requires an external dependency.
+        "manual",  # External dependency (through VulkanRuntime)
     ],
     deps = [
         ":VulkanRuntime",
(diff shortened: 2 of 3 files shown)


llvm-bazel/configure.bzl
```diff
@@ -10,4 +10,4 @@ load(":overlay_directories.bzl", "overlay_directories")
 OVERLAY_DIR = "llvm-project-overlay"
 
 def llvm_configure(name, overlay_path = OVERLAY_DIR, **kwargs):
-    overlay_directories(name=name, overlay_path=overlay_path, **kwargs)
+    overlay_directories(name = name, overlay_path = overlay_path, **kwargs)
```
chandlerc: I've seen this and some other minor Bazel fixes in a two PRs -- maybe factor it out as just cleanup?

llvm-bazel/llvm-project-overlay/mlir/BUILD
```diff
@@ -163,7 +163,8 @@ cc_library(
         "include/mlir-c/Bindings/Python/Interop.h",
     ],
     tags = [
-        "manual",  # TODO(gcmn): Add support for this target
+        "manual",  # External dependency
+        "nobuildkite",  # TODO(gcmn): Add support for this target
```
chandlerc, line 167: Should maybe have a canonical order and comment structure? (Differs from the above, not suggesting which one is best)

chandlerc (approved):
Just minor tweaks.

llvm-bazel/configure.bzl
GMNGeoffrey: Urgh yeah this is because my vscode extension for formatting build files (which will just operate on changed lines) has stopped working so I've been running `buildifier -r` on the command line. I can pull this out into a separate PR.

llvm-bazel/llvm-project-overlay/mlir/BUILD
GMNGeoffrey: Hmm I'm surprised buildifier isn't sorting them. These are different though. In this case the manual tag is there because this target uses an external dependency, so we want it to only get built when requested (to avoid pulling in that external dependency) even when we do get it to build. In some cases, the manual tag is only there because the target is currently broken, but once we get it working we should be able to remove it entirely.

GMNGeoffrey: Cleaned up the comments.

llvm-bazel/configure.bzl
GMNGeoffrey: Done in #50

PR #59, opened 2020-10-19
Introduce proper rule for changing the basename of a binary

GMNGeoffrey:
Bazel gets upset on Windows with this little hack, and honestly it's
right:

> in sh_binary rule @llvm-project//llvm:llvm-strip: Source file is a
Windows executable file, target name extension should match source file
extension


chandlerc (changes requested):
(as discussed offline, lets use the symlink feature)

GMNGeoffrey:
Reviewed internally, where I made the same change, so I'm going to submit before waiting for Chandler's official stamp.

PR #60, opened 2020-10-20
Configuration for a windows build

GMNGeoffrey:
This now builds with warnings-as-errors (which requires disabling quite
a few warnings).

I am by no means a Windows expert. I just got my hands on a Windows VM
and fixed things until everything built. Requires
https://github.com/google/llvm-bazel/pull/59,
https://reviews.llvm.org/D89760, and https://reviews.llvm.org/D89758
to build cleanly.

There are a lot of places where we set linkopts iff not windows. Might
want to factor that into a macro, but I decided to wait till we had more
configurations to see if that would remain the differentiator.

### DIFF
--- a/llvm-bazel/.bazelrc
+++ b/llvm-bazel/.bazelrc
@@ -63,6 +63,48 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
 build:generic_gcc --copt=-Werror --host_copt=-Werror
 
 ###############################################################################
+# Windows specific flags for building with VC.
+###############################################################################
+
+build:msvc --copt=/WX     # Treat warnings as errors...
+
+# ...but disable the ones that are violated
+build:msvc --copt=/wd4624 # destructor was implicitly defined as deleted
+build:msvc --copt=/wd4244 # possible loss of data
+build:msvc --copt=/wd4005 # macro redefinition
+build:msvc --copt=/wd4141 # inline used more than once
+build:msvc --copt=/wd4267 # initializing: possible loss of data
+build:msvc --copt=/wd4273 # multiple definitions with different dllimport
+build:msvc --copt=/wd4319 # zero-extending after complement
+build:msvc --copt=/wd4804 # comparisons between bool and int
+build:msvc --copt=/wd4805 # comparisons between bool and int
+
+build:msvc --linkopt=/WX # Treat warnings as errors...
+
+# ...but disable the ones that are violated.
+build:msvc --linkopt=/IGNORE:4001 # no object files
+build:msvc --linkopt=/IGNORE:4217 # mismatch import/export declspec
+
+# Enables unix-style runfiles link trees (requires symlink permission).
+# See: https://blogs.msvc.com/msvcdeveloper/2016/12/02/symlinks-msvc-10/
+# Generally: Enable Developer Mode in the Developer Settings page of the
+# system settings.
+build:msvc --experimental_enable_runfiles
+build:msvc --copt=/arch:AVX
+# Host and target are the same in msvc so don't waste time building both.
+build:msvc --distinct_host_configuration=false
+# Avoids incompatible versions of winsock and other badness.
+build:msvc --copt=/DWIN32_LEAN_AND_MEAN --host_copt=/DWIN32_LEAN_AND_MEAN
+# Why are min/max macros? No one knows.
+build:msvc --copt=/DNOMINMAX --host_copt=/DNOMINMAX
+# Yay for security warnings. Boo for non-standard.
+build:msvc --copt=/D_CRT_SECURE_NO_WARNINGS --host_copt=/D_CRT_SECURE_NO_WARNINGS
+# Necessary for M_* math constants.
+build:msvc --copt=/D_USE_MATH_DEFINES --host_copt=/D_USE_MATH_DEFINES
+
+###############################################################################
+
+###############################################################################
 # Configuration for building remotely using Remote Build Execution (RBE)
 # https://cloud.google.com/remote-build-execution/
 # Based on https://github.com/bazelbuild/bazel-toolchains/blob/master/bazelrc/bazel-1.0.0.bazelrc
--- a/llvm-bazel/.bazelrc
+++ b/llvm-bazel/.bazelrc
@@ -66,30 +66,20 @@ build:generic_gcc --copt=-Werror --host_copt=-Werror
 # Windows specific flags for building with VC.
 ###############################################################################
 
-build:msvc --copt=/WX     # Treat warnings as errors...
+build:msvc --copt=/WX --host_copt=/WX    # Treat warnings as errors...
 # ...but disable the ones that are violated
-build:msvc --copt=/wd4141 # inline used more than once
-build:msvc --copt=/wd4244 # conversion type -> type
-build:msvc --copt=/wd4267 # conversion size_t -> type
-build:msvc --copt=/wd4273 # multiple definitions with different dllimport
-build:msvc --copt=/wd4319 # zero-extending after complement
-build:msvc --copt=/wd4624 # destructor was implicitly defined as deleted
-build:msvc --copt=/wd4804 # comparisons between bool and int
-build:msvc --copt=/wd4805 # comparisons between bool and int
+build:msvc --copt=/wd4141 --host_copt=/wd4141 # inline used more than once
+build:msvc --copt=/wd4244 --host_copt=/wd4244 # conversion type -> type
+build:msvc --copt=/wd4267 --host_copt=/wd4267 # conversion size_t -> type
+build:msvc --copt=/wd4273 --host_copt=/wd4273 # multiple definitions with different dllimport
+build:msvc --copt=/wd4319 --host_copt=/wd4319 # zero-extending after complement
+build:msvc --copt=/wd4624 --host_copt=/wd4624 # destructor was implicitly defined as deleted
+build:msvc --copt=/wd4804 --host_copt=/wd4804 # comparisons between bool and int
+build:msvc --copt=/wd4805 --host_copt=/wd4805 # comparisons between bool and int
 
-build:msvc --linkopt=/WX # Treat warnings as errors...
+build:msvc --linkopt=/WX --host_linkopt=/WX # Treat warnings as errors...
 # ...but disable the ones that are violated.
-build:msvc --linkopt=/IGNORE:4001 # no object files
-
-# Enables unix-style runfiles link trees (requires symlink permission).
-# See: https://blogs.msvc.com/msvcdeveloper/2016/12/02/symlinks-msvc-10/
-# Generally: Enable Developer Mode in the Developer Settings page of the
-# system settings.
-# https://docs.bazel.build/versions/master/windows.html#enable-symlink-support
-build:msvc --enable_runfiles
-
-# Host and target are the same in msvc so don't waste time building both.
-build:msvc --distinct_host_configuration=false
+build:msvc --linkopt=/IGNORE:4001 --host_linkopt=/IGNORE:4001 # no object files
 
 # Yay for security warnings. Boo for non-standard.
 build:msvc --copt=/D_CRT_SECURE_NO_WARNINGS --host_copt=/D_CRT_SECURE_NO_WARNINGS
--- a/llvm-bazel/.bazelrc
+++ b/llvm-bazel/.bazelrc
@@ -67,40 +67,32 @@ build:generic_gcc --copt=-Werror --host_copt=-Werror
 ###############################################################################
 
 build:msvc --copt=/WX     # Treat warnings as errors...
-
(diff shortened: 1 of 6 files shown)


llvm-bazel/.bazelrc
```diff
@@ -62,6 +62,48 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
...
+# Host and target are the same in msvc so don't waste time building both.
+build:msvc --distinct_host_configuration=false
+# Avoids incompatible versions of winsock and other badness.
+build:msvc --copt=/DWIN32_LEAN_AND_MEAN --host_copt=/DWIN32_LEAN_AND_MEAN
+# Why are min/max macros? No one knows.
+build:msvc --copt=/DNOMINMAX --host_copt=/DNOMINMAX
+# Yay for security warnings. Boo for non-standard.
+build:msvc --copt=/D_CRT_SECURE_NO_WARNINGS --host_copt=/D_CRT_SECURE_NO_WARNINGS
+# Necessary for M_* math constants.
+build:msvc --copt=/D_USE_MATH_DEFINES --host_copt=/D_USE_MATH_DEFINES
```
chandlerc: I'd love to get some of the Windows Clang folks to help vet this...

Maybe @zmodem and @rnk could take a look?

chandlerc (commented):
The BUILD stuff makes sense.

There are a bunch of other non-Windows defines in the `config.h`, but they're only reached from code that is only compiled on non-Windows platforms, so no need to move them to Bazel really.

For the warnings and flags, I'd like to get more MSVC-aware folks to help comment.

Also might make sense to have a clang-cl config here as that is at least a thing on Windows.

llvm-bazel/.bazelrc
```diff
@@ -62,6 +62,48 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
...
+###############################################################################
+# Windows specific flags for building with VC.
+###############################################################################
+
+build:msvc --copt=/WX     # Treat warnings as errors...
+
+# ...but disable the ones that are violated
+build:msvc --copt=/wd4624 # destructor was implicitly defined as deleted
+build:msvc --copt=/wd4244 # possible loss of data
+build:msvc --copt=/wd4005 # macro redefinition
```
zmodem: Disabling the macro redefinition warning seems a bit surprising. I don't think we do that in the cmake or gn builds.

llvm-bazel/.bazelrc
```diff
@@ -62,6 +62,48 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
...
+# ...but disable the ones that are violated.
+build:msvc --linkopt=/IGNORE:4001 # no object files
+build:msvc --linkopt=/IGNORE:4217 # mismatch import/export declspec
+
+# Enables unix-style runfiles link trees (requires symlink permission).
+# See: https://blogs.msvc.com/msvcdeveloper/2016/12/02/symlinks-msvc-10/
+# Generally: Enable Developer Mode in the Developer Settings page of the
+# system settings.
+build:msvc --experimental_enable_runfiles
+build:msvc --copt=/arch:AVX
```
zmodem: We don't pass /arch:avx upstream, and I guess this will narrow the targets your builds will run on. As long as that's intentional, I suppose it's okay.

llvm-bazel/.bazelrc
```diff
@@ -62,6 +62,48 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
...
+
+# ...but disable the ones that are violated.
+build:msvc --linkopt=/IGNORE:4001 # no object files
+build:msvc --linkopt=/IGNORE:4217 # mismatch import/export declspec
+
+# Enables unix-style runfiles link trees (requires symlink permission).
+# See: https://blogs.msvc.com/msvcdeveloper/2016/12/02/symlinks-msvc-10/
+# Generally: Enable Developer Mode in the Developer Settings page of the
+# system settings.
+build:msvc --experimental_enable_runfiles
```
zmodem: I guess this is something Bazel specific? Should there be a comment about what it is? If users need to jump through hoops for this, at least it's nice to know what it's for.

llvm-bazel/.bazelrc
```diff
@@ -62,6 +62,48 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
...
+build:msvc --copt=/wd4273 # multiple definitions with different dllimport
+build:msvc --copt=/wd4319 # zero-extending after complement
+build:msvc --copt=/wd4804 # comparisons between bool and int
+build:msvc --copt=/wd4805 # comparisons between bool and int
+
+build:msvc --linkopt=/WX # Treat warnings as errors...
+
+# ...but disable the ones that are violated.
+build:msvc --linkopt=/IGNORE:4001 # no object files
+build:msvc --linkopt=/IGNORE:4217 # mismatch import/export declspec
```
zmodem: I would be interested to know why these two need to be disabled, especially the last one seems surprising to me.

llvm-bazel/.bazelrc
```diff
@@ -62,6 +62,48 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
...
+# Enables unix-style runfiles link trees (requires symlink permission).
+# See: https://blogs.msvc.com/msvcdeveloper/2016/12/02/symlinks-msvc-10/
+# Generally: Enable Developer Mode in the Developer Settings page of the
+# system settings.
+build:msvc --experimental_enable_runfiles
+build:msvc --copt=/arch:AVX
+# Host and target are the same in msvc so don't waste time building both.
+build:msvc --distinct_host_configuration=false
+# Avoids incompatible versions of winsock and other badness.
+build:msvc --copt=/DWIN32_LEAN_AND_MEAN --host_copt=/DWIN32_LEAN_AND_MEAN
```
zmodem: This doesn't get set in the cmake build. Typically code that includes windows.h and needs this set the macro in the source code.

Also, why are both --copt and --host_copt set here, but only --copt for the warnings further above?

llvm-bazel/.bazelrc
```diff
@@ -62,6 +62,48 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
...
+# Why are min/max macros? No one knows.
+build:msvc --copt=/DNOMINMAX --host_copt=/DNOMINMAX
```
zmodem: Same as for lean_and_mean, there are already defines of this around includes of windows.h; the cmake build doesn't set this macro globally.

llvm-bazel/.bazelrc
zmodem: I'm not familiar with _USE_MATH_DEFINES, but I don't see it set in the cmake build.

zmodem (commented):
The list of warnings looks reasonable to me except the one I commented on.

For reference, the cmake build disables more warnings:
https://github.com/llvm/llvm-project/blob/llvmorg-11.0.0/llvm/cmake/modules/HandleLLVMOptions.cmake#L531

And the gn build disabled fewer:
https://github.com/llvm/llvm-project/blob/llvmorg-11.0.0/llvm/utils/gn/build/BUILD.gn#L113

(Apologies if some of the guestions are weird; I don't know anything about Bazel.)

llvm-bazel/.bazelrc
GMNGeoffrey: Yes this is the comment above is saying. It enables symlink runfiles trees. It actually looks like this now exists as a non-experimental flag. Let me try that.

llvm-bazel/.bazelrc
GMNGeoffrey: It was needed because of the redefinition of LEAN_AND_MEAN :grin: Removed.

llvm-bazel/.bazelrc
```diff
@@ -62,6 +62,48 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
...
+build:msvc --linkopt=/IGNORE:4217 # mismatch import/export declspec
+
+# Enables unix-style runfiles link trees (requires symlink permission).
+# See: https://blogs.msvc.com/msvcdeveloper/2016/12/02/symlinks-msvc-10/
+# Generally: Enable Developer Mode in the Developer Settings page of the
+# system settings.
+build:msvc --experimental_enable_runfiles
+build:msvc --copt=/arch:AVX
+# Host and target are the same in msvc so don't waste time building both.
+build:msvc --distinct_host_configuration=false
```
GMNGeoffrey: @stellaraccident I'm somewhat skeptical of this option. Doesn't this mean that `-c dbg` will cause host tools to be rebuilt? https://docs.bazel.build/versions/master/guide.html#--distinct_host_configurationfalse

llvm-bazel/.bazelrc
GMNGeoffrey: Thanks. Done. Regarding `copt` vs `host_copt`, the `distinct_host_configuration=false` makes these unnecessary, but if we didn't have it then the warnings above would probably need to be disabled with `host_copt` as well.

llvm-bazel/.bazelrc
GMNGeoffrey: Done.

llvm-bazel/.bazelrc
GMNGeoffrey: Fixed. I think we inherited this from TF.

llvm-bazel/.bazelrc
GMNGeoffrey: For the purpose of making sure these build files are usable by dependent projects, probably better to leave it off. @stellaraccident can you comment on why we have this in IREE?

llvm-bazel/.bazelrc
GMNGeoffrey: 4001 is from having `cc_binary` with only deps and no srcs, I think. We get errors like:

> ERROR: C:/users/kbuilder/_bazel_kbuilder/paxle3de/external/llvm-project/llvm/BUILD:471:10: Linking of rule '@llvm-project//llvm:llvm-tblgen' failed (Exit 4001)
LINK : warning LNK4001: no object files specified; libraries used

4217 is not showing up for me now... Hmm I must have copy-pasta'd this from iree and then gotten it mixed up with 4273 when I was trimming things.

GMNGeoffrey:
Thanks for the review here! Really appreciate having some experienced eyes on this :-) I'm not really very familiar with windows dev. I stole a bunch of this from IREE and tried to trim out the cruft, but obviously missed some stuff. We have to hack around a bit more because we build TF.

GMNGeoffrey:
> Also might make sense to have a clang-cl config here as that is at least a thing on Windows.

I think that can be a later addition :-)

llvm-bazel/.bazelrc
zmodem: I guess I don't know what a symlink runfiles tree is, but maybe I don't need to know :) If it's not important, I'd recommend skipping it though, since requiring symlinks on Windows might not work for everyone.

llvm-bazel/.bazelrc
zmodem: Okay. 4001 sounds benign, but 4217 seems a bit scary so dropping that is probably good.

zmodem:
llvm-bazel/.bazelrc looks good to me now.

llvm-bazel/.bazelrc
GMNGeoffrey: It's a Bazel thing. For sandboxing purposes Bazel will create a symlink tree of files available to an action (e.g. a test) based on its dependencies. The action is then run with only its declared dependencies available. Bazel will use symlinks to create these trees. Let me see whether we can drop this.

GMNGeoffrey: Works without. Let's leave it off unless @stellaraccident is there a reason we need this in IREE that isn't just "TF uses it"?

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -209,10 +209,13 @@ cc_library(
     ],
     copts = llvm_copts,
     includes = ["include"],
-    linkopts = [
-        "-pthread",
-        "-ldl",
-    ],
+    linkopts = select({
```
r4nt, line 212: Should we move the common selects that are used multiple times into a starlark macro?

GMNGeoffrey: I commented on this in the description, but I think it's worth waiting to see how other platform configs shake out before refactoring this. It's not clear to me whether the current patterns (windows vs not-windows) will hold when we add additional configurations.

chandlerc (approved):
LGTM! Excited to see this go in, I want to get a Windows box set up myself so I can help debug this. Would be especially interested in a `clang-cl` config, but maybe that's something I or other folks interested can figure out...

PR #64, opened 2020-10-26
Integrate LLVM at llvm/llvm-project@cae4067ec1cd

GMNGeoffrey:
[cae4067ec1cd](https://github.com/llvm/llvm-project/commit/cae4067ec1cd)


chandlerc:
I'm a little uncomfortable with the zero-review merges with actual BUILD file updates.... do you actually want review? I was starting to look at this and noticed it had already been merged.

GMNGeoffrey:
I've been following the LLVM standard of submitting without review and getting post-commit review for simple changes. These are necessary to keep the build green, so I think merging them quickly is high-value. Most of these are just ported from internal patches that were authored by someone related to the commit they were for and seem non-controversial.

PR #70, opened 2020-10-27
Initial attempt at a MacOS port.

chandlerc:
This builds and produces a seemingly functional set of binaries. Only
tested on MacOS x86 laptop with XCode 12.

I've added some TODOs around parts that still seem brittle.

### DIFF
--- a/llvm-bazel/llvm-project-overlay/llvm/unittests/BUILD
+++ b/llvm-bazel/llvm-project-overlay/llvm/unittests/BUILD
@@ -465,6 +465,9 @@ cc_test(
         "$(STACK_FRAME_UNLIMITED)",
     ],
     linkstatic = 1,
+    tags = [
+        "local", # Not compatible with the sandbox on MacOS
+    ],
     visibility = ["//visibility:private"],
     deps = [
         "//llvm:AllTargetsCodeGens",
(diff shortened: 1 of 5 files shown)


GMNGeoffrey (approved):

chandlerc:
PTAL, had to mark a test as local, but now also passes all tests on Mac!

llvm-bazel/llvm-project-overlay/llvm/unittests/BUILD
```diff
@@ -465,6 +465,9 @@ cc_test(
         "$(STACK_FRAME_UNLIMITED)",
     ],
     linkstatic = 1,
+    tags = [
+        "local", # Not compatible with the sandbox on MacOS
+    ],
```
GMNGeoffrey, line 470: Just note that this appears to make the test take about 10 times as long. Would it be better to disable the gid/uid test entirely with `--gtest_filter`?

GMNGeoffrey (approved):

llvm-bazel/llvm-project-overlay/llvm/unittests/BUILD
chandlerc: I've checked and at least for me I'm not seeing a significant difference ... on my Linux machine, it takes about 8s to run the test just like on the buildkite instance? I think it's just cached vs. uncached.

I'd prefer this to skipping the test as the test granularity isn't super narrow -- we'd lose all permission management tests. =/

PR #71, opened 2020-10-27
Clean-up and remove hard-coded flag from `tblgen` Bazel rule.

chandlerc:
The hard coded flag disables the use of long string literals. This isn't
supported for all tablegen tools, and isn't desirable in any of them. It
is a workaround for compilers that can't handle the long string
literals. For compilers that can handle these literals, using them is
a (significant) compile time improvement.

I don't believe any compilers we're supporting with Bazel currently
require this flag, and if they do we should find out and pass it very
narrowly for those compilers. First step is to remove it here.

I've tested this both on a recent Clang and GCC 5 successfully.

### DIFF
--- a/llvm-bazel/llvm-project-overlay/llvm/tblgen.bzl
+++ b/llvm-bazel/llvm-project-overlay/llvm/tblgen.bzl
@@ -40,10 +40,10 @@ def gentbl(name, tblgen, td_file, td_srcs, tbl_outs, library = True, **kwargs):
             outs = [out],
             tools = [tblgen],
             message = "Generating code from table: %s" % td_file,
-            cmd = (("$(location %s) " + "-I external/llvm-project/llvm/include " +
+            cmd = (("$(location %s) -I external/llvm-project/llvm/include " +
                     "-I external/llvm-project/clang/include " +
-                    "-I $$(dirname $(location %s)) " + ("%s $(location %s) --long-string-literals=0 " +
-                                                        "-o $@")) % (
+                    "-I $$(dirname $(location %s)) " +
+                    "%s $(location %s) -o $@") % (
                 tblgen,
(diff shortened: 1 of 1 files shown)


GMNGeoffrey (approved):
Yeah this rule needs a lot of work :-)

ScottTodd:
This change appears to break builds on Windows (MSVC), at least on my machine. Originally noticed in IREE (https://github.com/google/iree/issues/3756):

```
ERROR: D:/tmp/bzlroot/3mtfpusj/external/llvm-project/llvm/BUILD:1665:16: C++ compilation of rule '@llvm-project//llvm:X86UtilsAndDesc' failed (Exit 2)
bazel-out/x64_windows-opt/bin/external/llvm-project/llvm/lib/Target/X86\X86GenInstrInfo.inc(37938): fatal error C1091: compiler limit: string exceeds 65535 bytes in length
Target @llvm-project//llvm:X86UtilsAndDesc failed to build
```

Reverting this change (`git revert c42f29eaab657c9e91e0a655e832e8804d234720`) fixes the build for me.

Here's my MSVC version:
```
λ cl
Microsoft (R) C/C++ Optimizing Compiler Version 19.25.28614 for x64
```

---

Found via `git bisect` ([docs](https://git-scm.com/docs/git-bisect)):

llvm_bisect_command.sh:
```bash
git submodule update
cd llvm-bazel
bazel build --config=msvc @llvm-project//llvm:X86UtilsAndDesc
```

console:
```bash
git bisect start --first-parent
git bisect bad
git bisect good f703e5ec90d
git bisect run bash -v D:\dev\scratch\llvm_bisect_command.sh
...
c42f29eaab657c9e91e0a655e832e8804d234720 is the first bad commit
```

PR #72, opened 2020-10-28
Add Clang and its unittests to the build.

chandlerc:
This covers both the libraries and main binaries for Clang, including
many of the test binaries and a few tools. Much like the LLVM `BUILD`,
it should be sufficient for users that are primarily interested in
linking and using the Clang libraries themselves and isn't sufficient to
build a complete toolchain.

I've also added the unittests which turned out to be very simple
(similar to the LLVM unittests). Again, this should very closely match
the experience of library users of Clang.

I've also built the `clang` executable and successfully compiled the
world's smallest test program or two, so it seems to be in working
order.

So far, only tested on Linux, but with both a Clang and GCC build.

### DIFF
--- a/llvm-bazel/llvm-project-overlay/clang/BUILD
+++ b/llvm-bazel/llvm-project-overlay/clang/BUILD
@@ -1,0 +1,1976 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+load("//llvm:tblgen.bzl", "gentbl")
+load("//llvm:binary_alias.bzl", "binary_alias")
+
+package(
+    default_visibility = ["//visibility:public"],
+    licenses = ["notice"],
+)
+
+exports_files([
+    "tools/clang-format/clang-format.el",
+    "tools/clang-format/clang-format-test.el",
+    "tools/clang-format/clang-format.py",
+    "tools/clang-rename/clang-rename.el",
+    "tools/extra/clang-include-fixer/tool/clang-include-fixer.el",
+    "tools/extra/clang-include-fixer/tool/clang-include-fixer-test.el",
+])
+
+cc_binary(
+    name = "clang-tblgen",
+    srcs = glob([
+        "utils/TableGen/*.cpp",
+        "utils/TableGen/*.h",
+    ]),
+    copts = [
+        "$(STACK_FRAME_UNLIMITED)",
+    ],
+    linkopts = ["-ldl"],
+    stamp = 0,
+    deps = [
+        "//llvm:Support",
+        "//llvm:TableGen",
+        "//llvm:config",
+    ],
+)
+
+gentbl(
+    name = "diagnostic_defs_gen",
+    tbl_outs = [(
+        "-gen-clang-diags-defs -clang-component=%s" % c,
+        "include/clang/Basic/Diagnostic%sKinds.inc" % c,
+    ) for c in [
+        "AST",
+        "Analysis",
+        "Comment",
+        "Common",
+        "CrossTU",
+        "Driver",
+        "Frontend",
+        "Lex",
+        "Parse",
+        "Refactoring",
+        "Sema",
+        "Serialization",
+    ]] + [
+        (
+            "-gen-clang-diag-groups",
+            "include/clang/Basic/DiagnosticGroups.inc",
+        ),
+        (
+            "-gen-clang-diags-index-name",
+            "include/clang/Basic/DiagnosticIndexName.inc",
+        ),
+    ],
+    tblgen = ":clang-tblgen",
+    td_file = "include/clang/Basic/Diagnostic.td",
+    td_srcs = glob(["include/clang/Basic/*.td"]),
+)
+
+gentbl(
+    name = "basic_arm_neon_inc_gen",
+    tbl_outs = [(
+        "-gen-arm-neon-sema",
+        "include/clang/Basic/arm_neon.inc",
+    )],
+    tblgen = ":clang-tblgen",
+    td_file = "include/clang/Basic/arm_neon.td",
+    td_srcs = [
+        "include/clang/Basic/arm_neon.td",
+        "include/clang/Basic/arm_neon_incl.td",
+    ],
+)
+
+gentbl(
+    name = "basic_arm_fp16_inc_gen",
+    tbl_outs = [(
+        "-gen-arm-neon-sema",
+        "include/clang/Basic/arm_fp16.inc",
+    )],
+    tblgen = ":clang-tblgen",
+    td_file = "include/clang/Basic/arm_fp16.td",
+    td_srcs = [
+        "include/clang/Basic/arm_fp16.td",
+        "include/clang/Basic/arm_neon_incl.td",
+    ],
+)
+
+gentbl(
+    name = "basic_arm_mve_aliases_gen",
+    tbl_outs = [(
+        "-gen-arm-mve-builtin-aliases",
+        "include/clang/Basic/arm_mve_builtin_aliases.inc",
+    )],
+    tblgen = ":clang-tblgen",
+    td_file = "include/clang/Basic/arm_mve.td",
+    td_srcs = [
+        "include/clang/Basic/arm_mve.td",
+        "include/clang/Basic/arm_mve_defs.td",
+    ],
+)
+
+gentbl(
+    name = "basic_arm_sve_builtins_gen",
+    tbl_outs = [(
+        "-gen-arm-sve-builtins",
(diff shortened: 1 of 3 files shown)


llvm-bazel/llvm-project-overlay/clang/BUILD
```diff
@@ -0,0 +1,1976 @@
...
+    ],
+    includes = ["include"],
+)
+
+# TODO: This should get replaced with something that actually generates the
+# correct version number.
+genrule(
+    name = "vcs_version_gen",
+    # This should be under lib/Basic, but because of how the include paths
+    # are passed through blaze, it's easier to drop generated files next to
```
GMNGeoffrey:
```suggestion
    # are passed through bazel, it's easier to drop generated files next to
```

llvm-bazel/llvm-project-overlay/clang/BUILD
```diff
@@ -0,0 +1,1976 @@
...
+        "lib/Basic/*.h",
+        "lib/Basic/Targets/*.cpp",
+        "lib/Basic/Targets/*.h",
+    ]),
+    hdrs = glob([
+        "include/clang/Basic/*.h",
+    ]),
+    copts = [
+        "-DHAVE_VCS_VERSION_INC",
+        "-Iexternal/llvm-project/clang/lib/Basic",
```
GMNGeoffrey, line 370: TODO: this is an awful hack :-D

llvm-bazel/llvm-project-overlay/clang/BUILD
```diff
@@ -0,0 +1,1976 @@
...
+)
+
+cc_library(
+    name = "index",
+    srcs = glob(
+        [
+            "lib/Index/*.cpp",
+            "lib/Index/*.h",
+        ],
+        # Broken upstream after r367615, and currently unused.
```
GMNGeoffrey: I can't find this revision

llvm-bazel/llvm-project-overlay/clang/BUILD
```diff
@@ -0,0 +1,1976 @@
...
+    tblgen = ":clang-tblgen",
+    td_file = "include/clang/Basic/arm_bf16.td",
+    td_srcs = [
+        "include/clang/Basic/arm_bf16.td",
+        "include/clang/Basic/arm_neon_incl.td",
+    ],
+)
+
+# We generate the set of builtin headers under a special subdirectory in the
+# 'bin' section of the blaze output so that they can be used as data
```
GMNGeoffrey:
```suggestion
# 'bin' section of the bazel output so that they can be used as data
```

llvm-bazel/llvm-project-overlay/clang/BUILD
```diff
@@ -0,0 +1,1976 @@
...
+        "//llvm:config",
+    ],
+)
+
+cc_library(
+    name = "c-bindings",
+    hdrs = glob(["include/clang-c/*.h"]),
+    deps = [
+        ":libclang",
+        #"//clang-tools-extra/clang-include-fixer:include-fixer-plugin",
```
GMNGeoffrey:
```suggestion
```

llvm-bazel/llvm-project-overlay/clang/BUILD
```diff
@@ -0,0 +1,1976 @@
...
+)
+
+cc_binary(
+    name = "libclang.so",
+    linkopts = [
+        "-nodefaultlibs",
+        "-lc_nonshared",
+        "-lgcc",
+        "-lgcc_eh",
+        "-Wno-unused-command-line-argument",
```
GMNGeoffrey, line 1659: These look like they won't work with MSVC

llvm-bazel/llvm-project-overlay/clang/BUILD
```diff
@@ -0,0 +1,1976 @@
...
+cc_library(
+    name = "clang-driver",
+    srcs = glob([
+        "tools/driver/*.cpp",
+        "tools/driver/*.h",
+    ]),
+    copts = [
+        # Disable stack frame size checks in the driver because
+        # clang::ensureStackAddressSpace allocates a large array on the stack.
+        "$(STACK_FRAME_UNLIMITED)",
```
GMNGeoffrey, line 1770: We use this basically everywhere else without comment. Is this usage significantly different or more interesting?

llvm-bazel/llvm-project-overlay/clang/include/clang/Config/config.h
```diff
@@ -0,0 +1,89 @@
+/* This generated file is for internal use. Do not include it from headers. */
```
GMNGeoffrey, line 1: Comment that this is a hand port? License header?

llvm-bazel/llvm-project-overlay/clang/include/clang/Config/config.h
```diff
@@ -0,0 +1,89 @@
...
+/* #undef CLANG_DEFAULT_STD_CXX */
+
+/* Default C++ stdlib to use. */
+#define CLANG_DEFAULT_CXX_STDLIB ""
+
+/* Default runtime library to use. */
+#define CLANG_DEFAULT_RTLIB ""
+
+/* Default unwind library to use. */
+#define CLANG_DEFAULT_UNWINDLIB ""
```
GMNGeoffrey, line 27: Is empty string valid for these?

llvm-bazel/llvm-project-overlay/clang/unittests/BUILD
```diff
@@ -0,0 +1,399 @@
...
+)
+
+cc_test(
+    name = "tooling_recursive_ast_visitor_tests",
+    size = "medium",
+    srcs = glob(["Tooling/RecursiveASTVisitorTests/*.cpp"]) + [
+        "Tooling/RecursiveASTVisitorTests/CallbacksCommon.h",
+    ],
+    copts = [
+        "-Iexternal/llvm-project/clang/unittests/Tooling",  # include "TestVisitor.h"
```
GMNGeoffrey: TODO: awful hack :grin:

GMNGeoffrey (approved):
LGTM with some nits.

llvm-bazel/llvm-project-overlay/clang/include/clang/Config/config.h
chandlerc: Yeah, this actually matches what the default build looks like with CMake. AFAIK, these are just here to enable customization of builds to specialize in some cases.

llvm-bazel/llvm-project-overlay/clang/include/clang/Config/config.h
chandlerc: There isn't really a great header for this upstream, but I've copied and adjusted the header used in the hand port of the config header from LLVM.

llvm-bazel/llvm-project-overlay/clang/BUILD
chandlerc: It is a bit more interesting. This code has a very specific bit of code that intentionally uses a large amount of stack and would trigger even a really reliable stack frame limit warning. Most of the other cases are somewhat incidental and maybe even debatable. But this one is really recognizing a very unusual bit of code for manipulating the stack in a weird way to cope with Linux kernel bugs. (See the comments above the mentioned function for all the wild details.)

llvm-bazel/llvm-project-overlay/clang/BUILD
chandlerc: True, but will any of this work with MSVC? I would assume we'd need a completely different strategy to build a .dll file here.

Mostly just unsure what the Windows port of this would best look like. Any suggestions on what I should do for now?

llvm-bazel/llvm-project-overlay/clang/BUILD
chandlerc: Yeah, I've just removed the whole thing as it doesn't seem necessary any more.

llvm-bazel/llvm-project-overlay/clang/BUILD
chandlerc: Sure.

llvm-bazel/llvm-project-overlay/clang/unittests/BUILD
chandlerc: Sure.

llvm-bazel/llvm-project-overlay/clang/BUILD
GMNGeoffrey: Yeah I don't know what a real windows port would look like. Just that in other places we've avoided diagnostic flags and such. Probably fine for now. I don't need clang, so my interest in making it build on windows is somewhat lower :grin:

GMNGeoffrey (approved):

PR #73, opened 2020-10-28
Integrate LLVM at llvm/llvm-project@23ed570af1cc

GMNGeoffrey:
[23ed570af1cc](https://github.com/llvm/llvm-project/commit/23ed570af1cc). Includes updates necessary for
[7a996027b984](https://github.com/llvm/llvm-project/commit/7a996027b984)

### DIFF
--- a/llvm-bazel/llvm-project-overlay/llvm/BUILD
+++ b/llvm-bazel/llvm-project-overlay/llvm/BUILD
@@ -193,7 +193,7 @@ cc_library(
         ]),
     }),
     hdrs = glob([
-        "include/llvm/Support/*.h",
+        "include/llvm/Support/**/*.h",
         "include/llvm/ADT/*.h",
     ]) + [
         "include/llvm/Support/VCSRevision.h",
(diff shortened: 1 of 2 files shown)


GMNGeoffrey:
This includes updates necessary for two separate commits that happened to be near each other so I manually stepped through history. I'm still trying to provide functioning build files for arbitrary commits, so would like to merge this with a rebase.

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -194,6 +194,7 @@ cc_library(
     }),
     hdrs = glob([
         "include/llvm/Support/*.h",
+        "include/llvm/Support/FileSystem/*.h",
```
GMNGeoffrey: Another option would be just globbing for `"include/llvm/Support/**/*.h"` which I think wouldn't be too unreasonable.

chandlerc: I'm good with either.

chandlerc (approved):
LGTM

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Decided to go with the globbier glob as likely to be more future-proof

PR #82, opened 2020-11-05
Remove the `includes` directive from `tblgen` libraries.

chandlerc:
This caused every single directory tree with a tblgen output in it to be
in the `-isystem` header search path. That is far from desirable, and
was hiding a large number of issues reaching both generated files and
source files with `#include`.

On MacOS it also manifested in an amazing problem where a header in
Clang's AST library called `Availability.h` ended up being used instead
of an internal header by the same name inside of the MacOS libc to
provide system availability macros to the libc headers. The resulting
errors were... fun.

Given that this requires us to put back a bunch of manual include
searches in `copts`, I've created a couple of macros to manage building
these. This way we can tweak the exact spelling or hook on selects as
needed.

This may *seem* to make things a lot worse, but in fact this makes
everything *much* more strict because so many wild header search paths
are now only narrowly introduced.


llvm-bazel/llvm-project-overlay/llvm/tblgen.bzl
```diff
@@ -28,11 +28,7 @@ def gentbl(name, tblgen, td_file, td_srcs, tbl_outs, library = True, **kwargs):
...
     if td_file not in td_srcs:
         td_srcs += [td_file]
-    includes = []
     for (opts, out) in tbl_outs:
-        outdir = out[:out.rindex("/")]
-        if outdir not in includes:
-            includes.append(outdir)
         rule_suffix = "_".join(opts.replace("-", "_").replace("=", "_").split(" "))
         native.genrule(
             name = "%s_%s_genrule" % (name, rule_suffix),
```
GMNGeoffrey, line 34: Of course GitHub won't let me comment on the actual line, but can you use your macros for the include arguments in `cmd` below?

GMNGeoffrey (approved):
Hooray for getting rid of spooky action at a distance. I wonder if we can use `strip_include_prefix` to help us out here? I think that should allow a stricter handling of includes that doesn't rely on the `include` argument.

llvm-bazel/llvm-project-overlay/llvm/tblgen.bzl
chandlerc: Not really... The command line flag syntax below I believe is subtly different from the C++ syntax. At the least you can't use the joined flag syntax (`-Ifoo`), and unsure how many other differences there might be. I'd much rather the macros I'm adding deal exclusively in compiler flags.

I think I can clean this up a bit though, i'll send a follow-up patch.

chandlerc:
> Hooray for getting rid of spooky action at a distance. I wonder if we can use `strip_include_prefix` to help us out here? I think that should allow a stricter handling of includes that doesn't rely on the `include` argument.

This is a great idea! I think you mentioned it to me before as well and I'm sorry I didn't follow up on it.

Unfortunately: https://github.com/bazelbuild/bazel/issues/12424

Fortunately, I think I can work around that. Bazel doesn't *really* use the distinction yet, and we can even future-proof against those features locally until this bug is fixed.

Anyways, I've completely removed the need for my hacky `-I` lines with this, **including all of the hacks in the target libraries!!!** That part took some ... doing. But I got it to work, and I think I now understand how to use this *everywhere*. I think this is now an even stronger improvement. I'm going to send a follow-up pull that remove the rest of the `-Iexternal/...` nonsense now that I know the right technique.

Of course, the target libraries really do have some awful things in them. I've built a construct that using `strip_include_prefix` to address this with a ... lot of complexity. I think it works though, and I've left comments to explain exactly what is going on and why.

PTAL (as this is essentially a totally different approach now).

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -453,13 +453,23 @@ cc_library(
...
     srcs = glob([
         "utils/TableGen/*.cpp",
-        "utils/TableGen/*.h",
         "utils/TableGen/GlobalISel/*.cpp",
+
+        # We have to include these headers here as well as in the `hdrs` below
+        # to allow the `.cpp` files to use file-relative-inclusion to find
+        # them, even though consumers of this library use inclusion relative to
+        # `utils/TableGen` with the `strip_includes_prefix` of this library.
+        # This mixture is likely incompatible with header modules.
```
GMNGeoffrey: "likely" incompatible? Seems like you turned it off so you concluded it was incompatible?

```suggestion
        # This mixture is incompatible with header modules.
```

Or are you not confident that there isn't some way to work around it and make it compatible?

```suggestion
        # This mixture appears to be incompatible with header modules.
```

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -2919,10 +2998,19 @@ gentbl(
...
+    name = "llvm-objcopy-headers",
+    hdrs = glob(["tools/llvm-objcopy/**/*.h"]),
+    strip_include_prefix = "tools/llvm-objcopy",
+)
+
 cc_binary(
     name = "llvm-objcopy",
     srcs = glob([
         "tools/llvm-objcopy/**/*.cpp",
+        # Note that we redundantly include the headers here to allow files to
```
GMNGeoffrey, line 3021: But here you don't need to turn off header_modules?

GMNGeoffrey (approved):
Nice! This looks to be heading in a good direction :-)

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: (it was the latter)

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: Ah, probably do here as well because they're duplicated, yeah.

chandlerc (commented):
Thanks for review, landing!

hlopko (commented):
Not nice workarounds, but I don't have a better solution other than fixing Bazel. LGTM.

PR #83, opened 2020-11-06
Remove most of the remaining include search hacks.

chandlerc:
Notably, all of the `-Iexternal/...` tricks are gone through the careful
application of extra entries in `srcs` to enable
file-relative-inclusion, and special header-only libraries which
re-expose header files with the desired prefixes using combinations of
`strip_include_prefix` and in one case `include_prefix`.

The result likely has some missing dependencies that will only be
detected if we enable header parsing in the future. I didn't want to try
to figure those out from scratch in every case without the tooling to
tell me if I'm wrong.

There are also still places where we should be doing intentional textual
headers, but cannot due to the Bazel bug that prevents
`strip_include_prefix` from working with them. I've left appropriate
comments about fixing those.

Even with the workarounds required here, I think this is a positive
step, and I don't really want to block on cleaning up all the code that
this copes with.

I also had to fix some of the build issues with the `llvm-exegesis`
stuff that I hadn't noticed before in order to test all of this, but
I did and that's included. The unittests and everything for that builds
now if you happen to have the correct stuff installed locally.

### DIFF
--- a/llvm-bazel/llvm-project-overlay/clang/BUILD
+++ b/llvm-bazel/llvm-project-overlay/clang/BUILD
@@ -330,11 +330,11 @@ cc_library(
         "include/clang/Basic/Version.inc",
         "include/clang/Config/config.h",
     ],
+    includes = ["include"],
     deps = [
         # We rely on the LLVM config library to provide configuration defines.
         "//llvm:config",
     ],
-    includes = ["include"],
 )
 
 # TODO: This should get replaced with something that actually generates the
@@ -346,6 +346,19 @@ genrule(
     # the other includes.
     outs = ["include/VCSVersion.inc"],
     cmd = "echo \"#define CLANG_REVISION \\\"git\\\"\" > $@",
+)
+
+# A hacky library to expose some internal headers of the `basic` library to its
+# own implementation source files using a stripped include prefix rather than
+# file-relative-inclusion. This is inherently non-modular as these headers will
+# be repeated in the sources below for file-relative-inclusion.
+cc_library(
+    name = "basic_internal_headers",
+    hdrs = glob([
+        "lib/Basic/*.h",
+    ]),
+    features = ["-header_modules"],
+    strip_include_prefix = "lib/Basic",
 )
 
 cc_library(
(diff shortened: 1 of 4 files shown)


GMNGeoffrey (approved):
Nice! Ah this is excellent because having these also constrains the name of the repository when used by others. Glad we got that taken care of :-)

Seems we should be able to use `strip_includes_prefix = ["include"]` instead of `includes = ["include"]` to get rid of those also. (not saying in this PR, just in general).

chandlerc:
Thanks, landing!

> Seems we should be able to use `strip_includes_prefix = ["include"]` instead of `includes = ["include"]` to get rid of those also. (not saying in this PR, just in general).

While we could, I don't think that's really an improvement, and would significantly increase the verbosity. Right now, not everything needs the `includes` thing. I'd actually like to remove it on all but the config libraries.

I guess maybe the `strip_includes_prefix` might be slightly cleaner in that it would *ensure* that every library did this, where as nothing really *ensures* that every library depends on the config one?

I'd somewhat like to ask the Bazel folks whether they have a preference about how this works or if there are hidden costs here.

GMNGeoffrey:
> Thanks, landing!
> 
> > Seems we should be able to use `strip_includes_prefix = ["include"]` instead of `includes = ["include"]` to get rid of those also. (not saying in this PR, just in general).
> 
> While we could, I don't think that's really an improvement, and would significantly increase the verbosity. Right now, not everything needs the `includes` thing. I'd actually like to remove it on all but the config libraries.
> 
> I guess maybe the `strip_includes_prefix` might be slightly cleaner in that it would _ensure_ that every library did this, where as nothing really _ensures_ that every library depends on the config one?
> 
> I'd somewhat like to ask the Bazel folks whether they have a preference about how this works or if there are hidden costs here.

Asking the Bazel folks SGTM. I think the advantage to the `strip_include_prefix` option is that it is a local change to the cc_library that affects how it is used by all consumers, whereas `includes` adds command line options to every library that depends on this one in a way that would affect its interactions with other libraries as well.

PR #84, opened 2020-11-07
Cleanup `buildifer` formatting and lint warnings.

chandlerc:
These aren't all trivial -- several of these showed questionable globs
that I've tried to fix in the way that seemed most effective.

---
This change is [Reviewable](https://reviewable.io/reviews/google/llvm-bazel/84)

### DIFF
--- a/llvm-bazel/llvm-project-overlay/clang/BUILD
+++ b/llvm-bazel/llvm-project-overlay/clang/BUILD
@@ -2,6 +2,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
 load("//llvm:tblgen.bzl", "gentbl")
 load("//llvm:binary_alias.bzl", "binary_alias")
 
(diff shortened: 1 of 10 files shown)


llvm-bazel/llvm-project-overlay/clang/BUILD
```diff
@@ -2,6 +2,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
```
GMNGeoffrey, line 5: Don't we need to add rules_cc to the workspace for this to work? https://github.com/bazelbuild/rules_cc#getting-started

It's not clear to me whether migrating to rules_cc is actually recommended right now. Internally I think we still use the native rules.

chandlerc: This works for me without anything else.

And `buildifier` has a lint complaint w/o it that this fixes. So for example, without this, VSCode thinks that all of our uses of the native rules are issues. It highlights them, etc etc.

I don't see how this can not be recommended with the `buildifier` tool linting it for the last several months?

GMNGeoffrey: Yeah I am also confused. I've filed https://github.com/bazelbuild/rules_cc/issues/92 about the documentation/tooling mismatch

chandlerc: Is there any harm in landing this to silence the lint warnings in the interim?

GMNGeoffrey: I have no idea :-D There's at least some indication that the future of rules_cc is unclear and not well documented: https://github.com/bazelbuild/rules_cc/issues/86. I'm a bit hesitant to move to it, given that

PR #90, opened 2020-11-13
Shard slow unit tests

GMNGeoffrey:
These tests are big and take a long time. If you have the appropriate
fanout, it's pretty painful to wait several minutes just because we've
globbed together a bunch of files. The `shard_count` is a pretty much
arbitrary reasonable-seeming choice.

Obviously not everyone has this fanout and sharding isn't free, so I
tried to stick to something not too large. Even without remote build
execution, 20 shards is likely to give you speedups on a reasonably
beefy machine. I also played with a couple other strategies: iterating
over globs to create one test per file, and determining the shard count
based on the number of files. The former doesn't really give us any of
the normal advantages of per-file test rules (which I would usually
prefer, but aren't feasible to maintain for a mostly-unsupported build
system). The latter didn't shard some of the bigger tests enough and we
still ended up with some very slow shards. Both of them also added to
the complexity of the build file.

As part of playing with this, I also increased our RBE worker pool to
150 workers, so increased the job count in the config to match. I can
send that as a separate PR if desirable.

### DIFF
--- a/llvm-bazel/.bazelrc
+++ b/llvm-bazel/.bazelrc
@@ -102,7 +102,7 @@ build:rbe --remote_instance_name=projects/llvm-bazel/instances/default_instance
 # this higher can make builds faster by allowing more jobs to run in parallel.
 # Setting it too high can result in jobs that timeout, however, while waiting
 # for a remote machine to execute them.
-build:rbe --jobs=50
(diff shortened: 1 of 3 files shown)


chandlerc (approved):
LGTM, this makes lots of sense.

PR #91, opened 2020-11-13
Move turnstyle timeout into the step

GMNGeoffrey:
Now possible with the introduction of `abort-after-seconds` in
https://github.com/softprops/turnstyle/pull/19. We've had a few
timeouts recently because the main branch got far behind `HEAD`.

### DIFF
--- a/.github/workflows/auto_integrate.yml
+++ b/.github/workflows/auto_integrate.yml
@@ -21,8 +21,6 @@ jobs:
     # Don't run on forks
     if: github.repository == 'google/llvm-bazel'
     runs-on: ubuntu-18.04
-    # Mostly so that busy-waiting through turnstyle is not indefinite.
-    timeout-minutes: 15
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/auto_integrate.yml
```diff
@@ -35,6 +33,7 @@ jobs:
         uses: softprops/turnstyle@v1
```
ScottTodd, line 33: pin to specific commit ref instead of release version?

ScottTodd (approved):

.github/workflows/auto_integrate.yml
GMNGeoffrey: No :-P Discussed offline, but the current security policy is just that we have to use a tagged release: https://opensource.google/docs/github/services/#actions

If the policy changes, we'll switch to a specific commit sha.

PR #103, opened 2020-12-01
Integrate LLVM at llvm/llvm-project@c21928285430

GMNGeoffrey:
[c21928285430](https://github.com/llvm/llvm-project/commit/c21928285430)

### DIFF
--- a/llvm-bazel/llvm_configs/abi-breaking.h.cmake
+++ b/llvm-bazel/llvm_configs/abi-breaking.h.cmake
@@ -20,7 +20,7 @@
 
 /* Allow selectively disabling link-time mismatch checking so that header-only
    ADT content from LLVM can be used without linking libSupport. */
-#if !LLVM_DISABLE_ABI_BREAKING_CHECKS_ENFORCING
+#if !defined(LLVM_DISABLE_ABI_BREAKING_CHECKS_ENFORCING) || !LLVM_DISABLE_ABI_BREAKING_CHECKS_ENFORCING
 
 // ABI_BREAKING_CHECKS protection: provides link-time failure when clients build
 // mismatch with LLVM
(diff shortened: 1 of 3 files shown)


llvm-bazel/llvm_configs/abi-breaking.h.cmake
```diff
@@ -20,7 +20,7 @@
 
 /* Allow selectively disabling link-time mismatch checking so that header-only
    ADT content from LLVM can be used without linking libSupport. */
-#if !LLVM_DISABLE_ABI_BREAKING_CHECKS_ENFORCING
+#if !defined(LLVM_DISABLE_ABI_BREAKING_CHECKS_ENFORCING) || !LLVM_DISABLE_ABI_BREAKING_CHECKS_ENFORCING
```
chandlerc, line 23: Do we not need to update our `abi-breaking.h` file?

chandlerc (changes requested):

llvm-bazel/llvm_configs/abi-breaking.h.cmake
GMNGeoffrey: We don't have one. We just have a Bazel rule that depends on the specific structure (but not this part of it)

https://github.com/google/llvm-bazel/blob/main/llvm-bazel/llvm-project-overlay/llvm/BUILD#L113-L124

chandlerc: Doh, sorry, forgot this one worked differently. Should have checked.

(Also sorry I didn't see your response earlier.)

chandlerc (approved):
LGTM

llvm-bazel/llvm_configs/abi-breaking.h.cmake
GMNGeoffrey: No worries

PR #112, opened 2020-12-11
Add a rough workaround for building `libclang.so` in a portable way.

chandlerc:
Sadly, there doesn't seem to be a terribly principled way to do this.
I've left some comments about the nature of this hack and the Bazel
issue tracking it. It was at least enough to unblock me building on
MacOS, no idea about Windows yet.


llvm-bazel/llvm-project-overlay/llvm/cc_plugin_binary.bzl
```diff
@@ -0,0 +1,44 @@
...
+forces the rule to statically link in its dependencies but to be linked as a
+shared "plugin" library. It then creates binary aliases to `.dylib` and
+`.dll` suffixed names for use on various platforms.
+"""
+
+load("@rules_cc//cc:defs.bzl", "cc_binary")
+load(":binary_alias.bzl", "binary_alias")
+
+def cc_plugin_binaries(name, **kwargs):
+    # Neither then ame of the plugin binary nor tags on whether it is built are
```
GMNGeoffrey: Nit: typo

```suggestion
    # Neither the name of the plugin binary nor tags on whether it is built are
```

llvm-bazel/llvm-project-overlay/llvm/cc_plugin_binary.bzl
```diff
@@ -0,0 +1,44 @@
...
+        **kwargs
+    )
+    binary_alias(
+        name = name + ".dll",
+        binary = ":" + name + ".so",
+    )
+    binary_alias(
+        name = name + ".dylib",
+        binary = ":" + name + ".so",
+    )
```
GMNGeoffrey: My concern here is that this macro creates rules that don't share its name and doesn't create a target that does share its name. That makes it difficult to use without reading the macro itself. You can't even `bazel build` it.

See https://docs.bazel.build/versions/master/skylark/bzl-style.html#macros for a discussion of macro style.

So in that sense I like the TF implementation better. However, to get the same result we don't need to go all the way to creating the same target multiple times. I think we don't even need to do a select. We could just do an extra alias


```suggestion
    cc_binary(
        name = name,
        linkshared = True,
        linkstatic = True,
        **kwargs
    )
    binary_alias(
        name = name + ".so",
        binary = ":" + name,
    )
    binary_alias(
        name = name + ".dll",
        binary = ":" + name,
    )
    binary_alias(
        name = name + ".dylib",
        binary = ":" + name,
    )
```

This makes the target buildable. It does still have the issue that building `name` doesn't build `name.so`, which the filegroup select has, so another option would be


```suggestion
    cc_binary(
        name = name + ".so",
        linkshared = True,
        linkstatic = True,
        **kwargs
    )
    binary_alias(
        name = name + ".dll",
        binary = ":" + name + ".so",
    )
    binary_alias(
        name = name + ".dylib",
        binary = ":" + name + ".so",
    )
    filegroup(
      name = name,
      srcs = [":" + name + select({
        "@bazel_tools//src/conditions:windows": ".dll",
        "@bazel_tools//src/conditions:darwin": ".dylib",
        "//conditions:default": ".so",
      })]
    )
```

GMNGeoffrey (approved):

llvm-bazel/llvm-project-overlay/llvm/cc_plugin_binary.bzl
chandlerc: Ah, I like it. I particularly like the combination which i've adopted. PTAL!

chandlerc (commented):
Really nice ideas.

I tried to commit the suggested edits, but it gave me a bunch of errors about invalid email address -- unsure if there is something you want to poke on your GH settings to fix that. But anyways, hopefully the result is a nice synthesis of these ideas.

There are still some more issues on MacOS that I need to debug, but this gets things (much) closer.

llvm-bazel/llvm-project-overlay/llvm/cc_plugin_library.bzl
```diff
@@ -0,0 +1,56 @@
...
+forces the rule to statically link in its dependencies but to be linked as a
+shared "plugin" library. It then creates binary aliases to `.dylib` and
+`.dll` suffixed names for use on various platforms.
+"""
+
+load("@rules_cc//cc:defs.bzl", "cc_binary")
+load(":binary_alias.bzl", "binary_alias")
+
+def cc_plugin_library(name, **kwargs):
+    # Neither the name of the plugin binary nor tags on whether it is built are
```
GMNGeoffrey, line 18: Comment needs updating

GMNGeoffrey (approved):
Weird about the email thing. I've definitely had suggested edits work before, so not sure what would've gone wrong.

llvm-bazel/llvm-project-overlay/llvm/cc_plugin_library.bzl
chandlerc: Updated, thanks. Merging, but lemme know if any further comment work would help.

PR #115, opened 2020-12-15
Integrate LLVM at llvm/llvm-project@68f53960e17d

GMNGeoffrey:
[68f53960e17d](https://github.com/llvm/llvm-project/commit/68f53960e17d).

This requires working around layering violations in `ExecutionEngine/`.
Here's a dot file graph showing the includes between directories in
`ExecutionEngine/`:
https://gist.github.com/GMNGeoffrey/2c9818d418a70517f8c628432e1c2447

### DIFF
--- a/llvm-bazel/llvm-project-overlay/llvm/BUILD
+++ b/llvm-bazel/llvm-project-overlay/llvm/BUILD
@@ -2049,16 +2049,26 @@ cc_library(
     ],
 )
 
+# There is some deep layering violation between these libraries and JITLink.
+# To workaround this, the same headers are included in multiple rules.
+cc_library(
+    name = "Orc_layering_violation_headers",
+    hdrs = glob([
+        "include/llvm/ExecutionEngine/Orc/*.h",
+        "include/llvm/ExecutionEngine/Orc/RPC/*.h",
+        "include/llvm/ExecutionEngine/JITLink/*.h",
+    ]),
+)
+
 cc_library(
     name = "OrcJIT",
     srcs = glob([
         "lib/ExecutionEngine/Orc/*.cpp",
         "lib/ExecutionEngine/Orc/*.h",
-        "lib/ExecutionEngine/Orc/RPC/*.cpp",
-        "lib/ExecutionEngine/OrcError/*.cpp",
     ]),
     hdrs = glob([
         "include/llvm/ExecutionEngine/Orc/*.h",
+        "include/llvm/ExecutionEngine/Orc/RPC/*.h",
     ]) + [
         "include/llvm-c/LLJIT.h",
         "include/llvm-c/Orc.h",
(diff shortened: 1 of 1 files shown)


llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -2049,16 +2049,26 @@ cc_library(
     ],
 )
 
+# There is some deep layering violation between these libraries and JITLink.
+# To workaround this, the same headers are included in multiple rules.
+cc_library(
+    name = "Orc_layering_violation_headers",
```
tpopp, line 2055: It's probably good to keep naming consistent, even if this is a special case.

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -2049,16 +2049,26 @@ cc_library(
...
     name = "OrcJIT",
     srcs = glob([
         "lib/ExecutionEngine/Orc/*.cpp",
         "lib/ExecutionEngine/Orc/*.h",
-        "lib/ExecutionEngine/Orc/RPC/*.cpp",
-        "lib/ExecutionEngine/OrcError/*.cpp",
     ]),
     hdrs = glob([
         "include/llvm/ExecutionEngine/Orc/*.h",
+        "include/llvm/ExecutionEngine/Orc/RPC/*.h",
```
tpopp, line 2071: Is this needed? I'm surprised that you took away `srcs` but need to add `hdrs`

tpopp (approved):
Looks fine. Let me know if you'd like me to take a look at this some time for additional cleanups. I can't promise I would dedicate much time to it, nor that I would do a better job than you, but I have at least had to do similar refactors before.

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Well the srcs paths don't exist anymore, which is why I deleted them. This isn't strictly necessary because it ends up getting these headers from `Orc_layering_violation_headers` via `OrcShared` and `OrcTargetProcess` and we don't have `layering_check` turned on, but there are direct includes between `Orc/` `Orc/RPC` going both ways. I think this is the rule these belong in.

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Yeah the naming in this file is a bit all over the place, but at least right now, weird workaround rules that don't correspond to any CMake libraries do have underscores. We could come up with a more consistent methodology for that. I kind of like verbose and ugly names for these kinds of verbose and ugly workarounds though :grin:

PR #120, opened 2020-12-16
Add missing BUILD rules for the OpenACC frontend library.

chandlerc:
These are only used by one unittest and the Flang frontend currently.
Somehow, the unittest in question compiles and runs successfully on all
the Linux hosts I've tested -- I really don't know how. My best guess is
that it finds the necessary generated header from an LLVM installation
that is made visible even through the sandbox to the build.

However, on MacOS, this doesn't work and the problem surfaces. Fix this
the right way by adding the missing build infrastructure matching the
OMP infrastructure (but simpler given the code simplicity of this
library).

With this, all of the LLVM and Clang unittests build and pass for me on
my Mac laptop.

### DIFF
--- a/llvm-bazel/llvm-project-overlay/llvm/unittests/BUILD
+++ b/llvm-bazel/llvm-project-overlay/llvm/unittests/BUILD
@@ -289,6 +289,7 @@ cc_test(
     srcs = glob(["FrontEnd/*.cpp"]),
     deps = [
         "//llvm:Analysis",
+        "//llvm:FrontendOpenACC",
         "//llvm:FrontendOpenMP",
         "//llvm:Passes",
         "//llvm:Support",
--- a/llvm-bazel/llvm-project-overlay/llvm/unittests/BUILD
+++ b/llvm-bazel/llvm-project-overlay/llvm/unittests/BUILD
@@ -286,7 +286,7 @@ cc_test(
 cc_test(
     name = "frontend_tests",
     size = "small",
-    srcs = glob(["FrontEnd/*.cpp"]),
+    srcs = glob(["Frontend/*.cpp"]),
     deps = [
         "//llvm:Analysis",
         "//llvm:FrontendOpenACC",
--- a/llvm-bazel/llvm-project-overlay/llvm/BUILD
+++ b/llvm-bazel/llvm-project-overlay/llvm/BUILD
@@ -1144,6 +1144,57 @@ cc_library(
     ],
 )
 
+filegroup(
+    name = "acc_td_files",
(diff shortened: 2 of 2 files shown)


GMNGeoffrey:
I think I have a simpler explanation. The build rule for these tests has a typo, `FrontEnd` vs `Frontend`. I have no idea what having a mac has to do with this. This points to a more general problem though. It would be nice to have a check that a glob expression actually returns any results...

GMNGeoffrey:
> It would be nice to have a check that a glob expression actually returns any results...

And indeed there is! https://docs.bazel.build/versions/3.3.0/be/functions.html#glob. We can set `allow_empty=False`. I'd prefer something that checks per include expression rather than the whole glob, but this should still help. Especially important for tests.

GMNGeoffrey:
Ah and to avoid the noise we can flip this flag `incompatible_disallow_empty_glob`: https://github.com/bazelbuild/bazel/issues/8195 instead. That actually checks that each glob expression is non-empty, which is nice. It also has the advantage that dependent projects can still happily use these build files, but we can enable stricter checks in our build bots.

GMNGeoffrey:
Sorry for quadruple posting. Turns out we have a *ton* of empty glob expressions and as I'm going through and taking them out, I think I've changed my mind. Most of these are harmless and future proofing: e.g. we grab all the cpp/h/def files in the include and lib directories. I'd rather not have to update these when someone adds a header. I'm back to thinking we should just add `allow_empty=False` to all our globs in test rules. If these are empty they will fail silently.

chandlerc:
As for why macos detects the issue: it's filesystem is (somewhat) case insensitive.

We should probably land this before the glob fix to keep bots green?

llvm-bazel/llvm-project-overlay/llvm/unittests/BUILD
```diff
@@ -289,6 +289,7 @@ cc_test(
     srcs = glob(["FrontEnd/*.cpp"]),
```
GMNGeoffrey: Let's include fixing this glob as part of this PR?

```suggestion
    srcs = glob(["Frontend/*.cpp"]),
```

GMNGeoffrey (approved):

GMNGeoffrey:
> As for why macos detects the issue: it's filesystem is (somewhat) case insensitive.

Aaaah

> We should probably land this before the glob fix to keep bots green?

Yeah for sure. I think it makes sense to fix this glob as part of this PR so it actually works on linux, so we can confirm these new build rules actually fix things. I'll add `allow_empty=False` in a followup.

PR #121, opened 2020-12-16
Disallow empty globs in tests

GMNGeoffrey:
Includes fixing one glob that was empty due to a typo. We don't want
silent do-nothing tests. There's also
`--incompatible_disallow_empty_glob`, but on balance I think it's better
to continue allowing empty globs in non-test rules as we have future
proofing there against source changes and this config is for source code
that gets updated without updating the build files.

See https://github.com/google/llvm-bazel/pull/120


chandlerc (approved):
LGTM (once the OpenACC patch is merged in).

PR #127, opened 2020-12-29
Integrate LLVM at llvm/llvm-project@930c74f

ScottTodd:
Rebased one commit from the `google-mlir` [branch](https://github.com/google/llvm-bazel/tree/google-mlir), updated the LLVM submodule, and manually fixed the `mlir-spirv-cpu-runner` target.


GMNGeoffrey:
Need to update in the `mlir-spirv-cpu-runner` rule which is not present internally for some reason (maybe we just delete it for now?)

GMNGeoffrey (approved):

PR #128, opened 2020-12-30
Disallow more empty test globs

GMNGeoffrey:
Apparently I messed up my regex and missed a bunch of these in
https://github.com/google/llvm-bazel/pull/121/

### DIFF
--- a/llvm-bazel/llvm-project-overlay/clang/unittests/BUILD
+++ b/llvm-bazel/llvm-project-overlay/clang/unittests/BUILD
@@ -10,10 +10,13 @@ package(
 cc_test(
     name = "ast_tests",
     size = "medium",
-    srcs = glob([
-        "AST/*.cpp",
-        "AST/*.h",
-    ]),
+    srcs = glob(
+        [
+            "AST/*.cpp",
+            "AST/*.h",
+        ],
+        allow_empty = False,
+    ),
     shard_count = 20,
     deps = [
         "//clang:ast",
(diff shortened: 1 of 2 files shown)


GMNGeoffrey:
@rupprecht if you could take a look

rupprecht:
It doesn't look like buildozer has a command that can set attributes like this, so regex it is.

Not sure what regex you're using, but I think this should be safe to automate, if you'd like to try this for more BUILD files:

`sed -i '/allow_empty = False/d; s/glob(/glob(allow_empty=False,/'  clang/unittests/BUILD && buildifier -a -v`

GMNGeoffrey:
> It doesn't look like buildozer has a command that can set attributes like this, so regex it is.
> 
> Not sure what regex you're using, but I think this should be safe to automate, if you'd like to try this for more BUILD files:
> 
> `sed -i '/allow_empty = False/d; s/glob(/glob(allow_empty=False,/' clang/unittests/BUILD && buildifier -a -v`

It's a bit more complicated because include is a positional argument and has to come first. I think I screwed up the multiline match the first time I did it. I got one that worked now though

rupprecht:
> It's a bit more complicated because include is a positional argument and has to come first

Making the substitution `s/glob(/glob(allow_empty=False,/` indeed breaks that, but running buildifier should restore it to the right place IIUC. However, it doesn't remove duplicate `allow_empty = False` entries, so the initial `/allow_empty = False/d;` is needed to clear that out.

rupprecht (approved):

GMNGeoffrey:
> > It's a bit more complicated because include is a positional argument and has to come first
> 
> Making the substitution `s/glob(/glob(allow_empty=False,/` indeed breaks that, but running buildifier should restore it to the right place IIUC. However, it doesn't remove duplicate `allow_empty = False` entries, so the initial `/allow_empty = False/d;` is needed to clear that out.

Ah you're right, buildifier does fix that, nice! I think your other command would delete the whole line, which could have other stuff, but `s/allow_empty=False,\?//` should do it :-)

PR #130, opened 2021-01-07
Work around https://github.com/bazelbuild/bazel/issues/3828

scentini:
Current Bazel has a bug (https://github.com/bazelbuild/bazel/issues/3828) when using `strip_include_prefix` that causes Bazel to erroneosly report a missing dependency declaration.

This bug shows up when trying to `bazel build @llvm-project//llvm:NVPTXCodeGen`:
```
 undeclared inclusion(s) in rule '@llvm-project//llvm:NVPTXCodeGen':
this rule is missing dependency declarations for the following files included by 'external/llvm-project/llvm/lib/Target/NVPTX/NVPTXAsmPrinter.cpp':
  'external/llvm-project/llvm/lib/Target/NVPTX/TargetInfo/NVPTXTargetInfo.h'
```

This has been fixed in https://github.com/bazelbuild/bazel/commit/e3b7e17b05f13ff183a4d7efec8ec797f3f5eaa3. This PR is a workaround until we have a Bazel version that contains the fix.


scentini:
cc @mihaimaruseac

llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -1730,6 +1730,7 @@ gentbl(
         srcs = ["lib/Target/" + target["name"] + "/TargetInfo/" + target["name"] + "TargetInfo.cpp"],
         hdrs = glob(["lib/Target/" + target["name"] + "/TargetInfo/*.h"]),
         copts = llvm_copts,
+        includes = ["lib/Target/" + target["name"]],
```
GMNGeoffrey, line 1722: Please add a TODO to remove this when the Bazel issue is fixed. Any idea what release that will be?

```suggestion
        # Workaround for https://github.com/bazelbuild/bazel/issues/3828
        # TODO(gcmn): Remove this when upgrading to a Bazel version containing
        # https://github.com/bazelbuild/bazel/commit/e3b7e17b05f1
        includes = ["lib/Target/" + target["name"]],
```

GMNGeoffrey (commented):
Thanks for the patch!

Are you able to reproduce this issue just with 

`bazel build @llvm-project//llvm:NVPTXCodeGen` ?

I can't on Linux Debian, although I have heard reports from people building on Windows. Note that I'm building with `--config=generic_clang`. At least some config is required, since that sets the C++ standard and such and I don't think there's anything in `generic_clang` that would somehow fix this issue.

mihaimaruseac:
I am able to reproduce on the Linux work desktop using that build command. Will apply patch manually and validate, after a the remaining meetings for today.

Thank you very much for the patch.

GMNGeoffrey:
> I am able to reproduce on the Linux work desktop using that build command. Will apply patch manually and validate, after a the remaining meetings for today.
> 
> Thank you very much for the patch.

Huh. What config are you using? I don't get the issue with either `generic_clang` or `generic_gcc`. I'm curious what's different about your setup vs mine, since I think we're running the exact same OS. Are you building directly in llvm-bazel itself or through TF?

mihaimaruseac:
This seems to make `bazel build @llvm-project//llvm:NVPTXCodeGen` work on my system. Trying to build TF now.

@GMNGeoffrey it is through TF, via Bazel. I can send more details via chat.

GMNGeoffrey:
> This seems to make `bazel build @llvm-project//llvm:NVPTXCodeGen` work on my system. Trying to build TF now.
> 
> @GMNGeoffrey it is through TF, via Bazel. I can send more details via chat.

Ok well I don't understand *why* that would make a difference, but that at least explains what's different about how you're building :-D

@chandlerc, is adding these includes back in going to mess up the build on OSX? Can you test it out?

mihaimaruseac:
Any updates? I cannot apply the patch via Bazel patching when fetching the repository :(

GMNGeoffrey:
@chandlerc waiting for you input here

> Any updates? I cannot apply the patch via Bazel patching when fetching the repository :(

Why not?

mihaimaruseac:
```
ERROR: An error occurred during the fetch of repository 'llvm-project-raw':
   Traceback (most recent call last):
        File "/tmp/tf/cp/tf/third_party/repo.bzl", line 107, column 25, in _tf_http_archive
                _apply_patch(ctx, ctx.attr.patch_file)
        File "/tmp/tf/cp/tf/third_party/repo.bzl", line 65, column 14, in _apply_patch
                ctx.patch(patch_file, strip = 1)
Error in patch: Error applying patch /tmp/tf/cp/tf/third_party/llvm-project/patch.patch: Cannot find file to patch (near line 3), old file name (llvm-project-overlay/llvm/BUILD) doesn't exist, new file name (llvm-project-overlay/llvm/BUILD) doesn't exist.
ERROR: no such package '@llvm-project-raw//': Error applying patch /tmp/tf/cp/tf/third_party/llvm-project/patch.patch: Cannot find file to patch (near line 3), old file name (llvm-project-overlay/llvm/BUILD) doesn't exist, new file name (llvm-project-overlay/llvm/BUILD) doesn't exist.
```

mihaimaruseac:
Wait, the problem was on my side, I was patching the wrong project. Seems to be working now.

mihaimaruseac:
So I think we no longer need this as I can patch TF. In fact, it seems we need to patch in several more places, depending on environment (Windows build fails on a different place)

GMNGeoffrey:
> So I think we no longer need this as I can patch TF. In fact, it seems we need to patch in several more places, depending on environment (Windows build fails on a different place)

If this is indeed a fix for a Bazel bug, we should fix it here. I just want to check that it doesn't break other things in the process

mihaimaruseac:
Bazel 4.0 that has been released in the past 24 hours no longer has the issue (it has the proper fix inside Bazel)

mihaimaruseac:
Actually it seems Bazel 4.0 still has the same issue, after testing with the TF patch removed

GMNGeoffrey:
> Actually it seems Bazel 4.0 still has the same issue, after testing with the TF patch removed

I don't see anything about this fix in the release notes. Maybe it's not included?

GMNGeoffrey:
@scentini can you comment on whether this is supposed to be in Bazel 4.0?

scentini:
No, the 4.0 baseline is a couple of days older than the fix, we'll need to wait for the next release.

GMNGeoffrey (approved):
Haven't heard from Chandler. I'm going to merge this.

PR #134, opened 2021-01-12
[MLIR] Update build for StandardOps/Utils

rsuderman:
Some files were missing from the bazel BUILD file for StandardOps.


GMNGeoffrey:
Please also bump the LLVM submodule to the appropriate commit. Those directories don't exist at the current LLVM commit on main

PR #150, opened 2021-02-11
Integrate LLVM at llvm/llvm-project@418c21

rsuderman:
[418c21](https://github.com/llvm/llvm-project/commit/418c218efa95)


GMNGeoffrey:
Ah lemme fix the config diff thing. One sec

rsuderman:
I thought I just handled it in the new push but apparently not?

GMNGeoffrey:
> I thought I just handled it in the new push but apparently not?

We need to actually update the configs to properly set that :-) The test is literally just a change detector to say "hey this thing changed"

GMNGeoffrey (approved):
Okie dokie. Our current runners do not have `mallinfo2`, so we don't want to set that. At some point, of course, we should make this actually configurable, but this is right for now

PR #156, opened 2021-02-23
Initial implementation of gentbl as a Starlark rule

GMNGeoffrey:
This should avoid having to list all transitive dependencies in every
invocation of gentbl and enable much nicer includes specification. This
is designed to allow a drop-in replacement to ease transition to the
new functionality. All existing usages of gentbl should also work with
this rule. I tested out fully transitioning the current build rules to
use td_library, but am going to submit this as just the change to
the rule, with the BUILD file changes coming later to make things more
incremental.


chandlerc:
I would really suggest getting @hlopko to review (can't properly @ him here for some reason). He understands this part of Bazel and Starlark much better than I do.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -4,7 +4,128 @@
 
 """BUILD extensions for MLIR table generation."""
 
-def gentbl(name, tblgen, td_file, tbl_outs, td_srcs = [], td_includes = [], td_relative_includes = [], strip_include_prefix = None, test = False):
+# A provider with one field, transitive_sources.
```
hlopko: Nits: The comment is stale, and you could use doc attribute to specify documentation in a way that can be later extracted by stardoc (https://docs.bazel.build/versions/master/skylark/lib/globals.html#parameters-23)

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -4,7 +4,128 @@
...
+TdFiles = provider(fields = ["transitive_sources", "transitive_includes"])
+
+def _get_dep_transitive_srcs(dep):
+    if TdFiles in dep:
+        return dep[TdFiles].transitive_sources
+    return dep[DefaultInfo].files
```
hlopko, line 26: I'd vote for not allow depending on filegroups now that there is a real rule with its own provider - that's why you look into DefaultInfo, to support filegroups right? DefaultInfo is provided by so many rules and nothing would stop me from depending on something wrong like cc_binary. Wdyt?

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -4,7 +4,128 @@
...
+    trans_includes = _get_transitive_includes(ctx.attr.includes, ctx.attr.deps)
+    return [TdFiles(
+        transitive_sources = trans_srcs,
+        transitive_includes = trans_includes,
+    )]
+
+td_library = rule(
+    _td_library_impl,
+    attrs = {
+        "srcs": attr.label_list(allow_files = True),
```
hlopko, line 110: Can `srcs` be empty? If not, `allow_empty` = False`

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -4,7 +4,128 @@
...
+        "includes": attr.string_list(),
+        "deps": attr.label_list(),
```
hlopko: Depending on your opinion on depending on filegroup we could add a restriction that only rules providing TdFiles can go into deps. Also for gentbl_rule.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -4,7 +4,128 @@
...
+    td_file = ctx.file.td_file
+    srcs = []
+    srcs.extend(ctx.files.td_srcs)
+    if td_file not in srcs:
+        srcs.append(td_file)
+
+    trans_srcs = _get_transitive_srcs(srcs, ctx.attr.deps)
+    trans_includes = _get_transitive_includes(ctx.attr.td_includes, ctx.attr.deps)
+
+    td_includes_cmd = ["-I%s" % (include,) for include in trans_includes.to_list()]
```
hlopko: `depset.to_list()` is frowned upon :) Flattening the depset can be source of slowness in larger builds. Let's not trigger djasper@ :) Use https://docs.bazel.build/versions/master/skylark/lib/Args.html and https://docs.bazel.build/versions/master/skylark/lib/Args.html#add_all.

Instead of `["-I%s" % (include,) for include in trans_includes.to_list()]`
you could write `args.add_all(trans_includes, format_each = "-I%s")`

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -4,7 +4,128 @@
...
+    output = "-o=%s" % (ctx.outputs.output.path,)
+
+    args = ctx.attr.opts + [
+        td_file.path,
+        "-I=%s" % (ctx.genfiles_dir.path,),
+    ] + td_includes_cmd + [output]
+
+    ctx.actions.run(
+        outputs = [ctx.outputs.output],
+        inputs = trans_srcs.to_list(),
```
hlopko: No need to run to_list there

hlopko (approved):
Looks really good! Just a few comments. Generally LGTM after to_list() is fixed.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
GMNGeoffrey: Yeah that was the intent and I went back and forth on whether that was a good idea. When I was moving over all the rules in this file it made things a bit easier because I had an intermediate state that worked. I think it might also ease the rollout because someone won't have to switch over all their dependencies in order to adopt the new format.

What about leaving filegroups allowed here for now with a TODO to remove them once a reasonable transition period has passed?

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
GMNGeoffrey: A rule with only deps seems reasonable, no?

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
GMNGeoffrey: I think it makes sense to do that for gentbl_rule regardless (which I should probably rename, actually) since deps was not a pre-existing attribute. For this one, see my comment above.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -20,72 +141,22 @@ def gentbl(name, tblgen, td_file, tbl_outs, td_srcs = [], td_includes = [], td_r
...
-        ]
-    for td_include in td_relative_includes:
-        td_includes_cmd += [
-            "-I%s/%s" % (native.package_name(), td_include),
-            "-I$(GENDIR)/%s/%s" % (native.package_name(), td_include),
-        ]
-
-    local_inc = "-I $$(dirname $(location %s))" % td_file
-
-    if test:
```
GMNGeoffrey, line 45: @joker-eph can you comment on what the use case for this target is?

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
hlopko: Your call, TODO sgtm.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
hlopko: (I have no clue what tblgen is so maybe it does, I wouldn't know :)

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
hlopko: That makes good sense!

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -27,19 +29,33 @@ def _get_transitive_srcs(srcs, deps):
...
     )
 
 def _get_transitive_includes(includes, deps):
     return depset(
-        includes,
+        direct = includes,
         transitive = [_get_dep_transitive_includes(dep) for dep in deps],
     )
 
+def _process_includes(ctx, raw_includes):
```
hlopko: I wonder if there's a more intention revealing name, wdyt about resolve_includes?

hlopko (approved):
Looks great, just a nit.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
GMNGeoffrey: Sure. I'm also trying to find a better way to get the current package path and current repository path. `native.package_name()` and `native.repository_name()` can only be called during loading. Basically I want to be able to replicate the includes parameters from `cc_library` but AFAICT that isn't actually implemented in Starlark. Like `strip_include_prefix` treats absolute paths as relative to the repository (https://docs.bazel.build/versions/master/be/c-cpp.html#cc_library.strip_include_prefix), but I can't figure out how to do that. Is there something that has this functionality implemented in Starlark?

GMNGeoffrey: Ok similarly, the way I have this implemented doesn't work. I'm testing this out in IREE and the package_path I'm getting is just `llvm-project/mlir/` as opposed to `external/llvm-project/mlir/`. How do I get the actual path here? The documentation for root says that they are "are mapped together into a single directory tree to form the execution environment"
(https://docs.bazel.build/versions/master/skylark/lib/root.html), which suggests to me that I shouldn't have to be doing all this futzing with various paths, but that doesn't seem to be the case.

GMNGeoffrey: I came up with the solution of adding the BUILD file as a private dep in the rule

```bzl
"_build_file": attr.label(default = ":BUILD", allow_single_file = True),
```

is that the best we've got?

GMNGeoffrey: I found `ctx.label`. I still feel like I'm doing a silly amount of path munging here.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -1,10 +1,175 @@
...
+
+    args = ctx.actions.args()
+    args.add_all(ctx.attr.opts)
+    args.add(td_file)
+    args.add("-I", ctx.genfiles_dir.path)
+    args.add("-I", td_file.dirname)
+    args.add_all(trans_includes, before_each = "-I")
+
+    # Can't use map_each because we need ctx.genfiles_dir and map_each can't be
+    # a closure.
```
hlopko: Super nit: I'd remove the comment

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -1,10 +1,175 @@
...
+    package = ctx.label.package
+    workspace_root = ctx.label.workspace_root
+    workspace_root = workspace_root if workspace_root else "."
+    resolved_includes = []
+    for include in includes:
+        if not include.startswith("/"):
+            include = "/" + package + "/" + include
+        include = workspace_root + include
+        resolved_includes.append(include)
+        resolved_includes.append(ctx.genfiles_dir.path + "/" + include)
```
hlopko: Only FYI: there's also ctx.bin_dir, which unless a bazel flag is used, is the same as ctx.genfiles_dir. By default ctx.actions.declare_file puts the file in the bin_dir. Sadly, internally in Google they are not the same. Just so you know in case you are wondering why some file is in genfiles in bazel but it's not in blaze.

hlopko (approved):

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
GMNGeoffrey: It's a breadcrumb for future people who are like me and would try to make this a single `add_all` with a map_each expanding to two include paths. I think it's worth saving them the effort.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
GMNGeoffrey: Yes I encountered this when porting to Blaze, which is why I added `output_to_genfiles` to this rule so I get consistent behavior in both. Does that seem like the correct workaround? I think only things in genfiles can be used by later rules, right? So since this is creating header files and such to be included later I do want it to output to genfiles (although as you note, by default in Bazel there's no difference):

> output_to_genfiles
> default = False
If true, the files will be generated in the genfiles directory instead of the bin directory. Unless you need it for compatibility with existing rules (e.g. when generating header files for C++), do not set this flag.

https://docs.bazel.build/versions/master/skylark/lib/globals.html#rule.output_to_genfiles

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
hlopko: Oh right I didn't realize that the intent was to have one add_all call. Then the comment makes sense you're right. Thanks for explaining.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
hlopko: I'm not sure I follow "I think only things in genfiles can be used by later rules, right?". Typically everything works fine when everything uses binfiles. If some rule uses genfiles, for example it generates a C++ source file there, then other rules need to make sure they work with genfiles. In the generated C++ source file example it would mean to correctly pass `-I` or `-iquote` for both genfiles and binfiles (and non-generated source files too).

hlopko: So maybe at the end the problem will be solved by adding 3 includes by default - `include`, `ctx.genfiles_dir.path + "/" + include`, and `ctx.binfiles_dir.path + "/" + include`.

hlopko: Alternatively, we could not remember include directories in the provider, but all the possible included files? And then map files to their parent directories when creating args. Am I making sense? I am rushing this reply on a Friday afternoon :)

hlopko (approved):

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
GMNGeoffrey: Ah, that documentation made me thing that maybe if you were generating a C++ header you needed to make sure it output to genfiles. Yeah perhaps we need to add all three.

> Alternatively, we could not remember include directories in the provider, but all the possible included files? And then map files to their parent directories when creating args.

Well we already have all possible included files (that's the deps), but we still need to know the include search paths because the way in which they're included is going to be relative to those, not just by their basename. That is, we need to add the `mlir/include` directory to the include paths so that we can reference `mlir/include/mlir/IR/OpBase.td` as `mlir/IR/OpBase.td`. This is basically the same as how the C++ compiler works (which is why it would be super helpful if there was a Starlark implementation of that I could reference).

GMNGeoffrey: Ok yeah, there are a number of rules that expect generated outputs to be generated in genfiles (since that's where genrule puts them). I think it may still be good to make this rule more tolerant and have its inclusion paths prefixed with all three, but also I think it needs to output to genfiles.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
GMNGeoffrey: Chatted in another forum.

PR #162, opened 2021-03-03
Adopt td_library

GMNGeoffrey:
Following https://github.com/google/llvm-bazel/pull/156, this avoids
needing to list all transitive include dependencies and tracks include
directories.


GMNGeoffrey:
Note that these changes were already reviewed internally

ThomasRaoux (approved):

PR #171, opened 2021-03-10
Add support for mingw-gcc.

rubensf:
Note that this is also blocked by #130.
That error also affects @llvm-project/llvm:llvm-mca-headers.

This doesn't build _everything_. Particularly, the last
error I got through was on setenv not being declared in
llvm/llvm/unittests/Support/ThreadPool.cpp. I imagine the issue might be
related to some of the other headers that we're including for linux and
posix only.

Note that I can't get mingw-gcc to work with RBE. GCC on windows
will try to canonize paths that look shorter than the original ones:
https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/libcpp/files.c#L405.

That means that relative includes (eg "lib/Support/../../", eg [1]) are made
absolute, which end up showing an absolute path on the `.d` files.

[1] https://github.com/llvm/llvm-project/blob/main/llvm/lib/Support/AArch64TargetParser.cpp#L58

### DIFF
--- a/llvm-bazel/.bazelrc
+++ b/llvm-bazel/.bazelrc
@@ -73,6 +73,7 @@ build:generic_gcc --copt=-Werror --host_copt=-Werror
 ###############################################################################
 
 build:mingw_gcc --config=generic_gcc
+build:mingw_gcc --compiler=mingw-gcc
 
 # Other windows toolchains have these by default
 build:mingw_gcc --cxxopt=-DHAVE_UNISTD_H=1 --host_cxxopt=-DHAVE_UNISTD_H=1
--- a/llvm-bazel/.bazelrc
+++ b/llvm-bazel/.bazelrc
@@ -77,6 +77,18 @@ build:mingw_gcc --config=generic_gcc
 # Other windows toolchains have these by default
 build:mingw_gcc --cxxopt=-DHAVE_UNISTD_H=1 --host_cxxopt=-DHAVE_UNISTD_H=1
 
+# The following link options should go on the BUILD rules instead, but:
+# 1. It's very challenging to make select constraints work properly
+# 2. Trying to set them for windows in general will trigger msvc warnings for
+# unknown flags.
+
+# Needed by @llvm-project//llvm:llvm-tblgen, @llvm-project//clang:clang-tblgen, @llvm-project//clang:clang-scan-deps
+build:mingw_gcc --linkopt=-lole32 --host_linkopt=-lole32
+# Needed by @llvm-project//llvm:llvm-tblgen, @llvm-project//clang:clang-tblgen, @llvm-project//clang:clang-scan-deps
+build:mingw_gcc --linkopt=-luuid --host_linkopt=-luuid
+# Needed by @llvm-project//clang:clang-scan-deps
+build:mingw_gcc --linkopt=-lversion --host_linkopt=-lversion
+
 # Needed to avoid  "File too big" error. -Wa,-mbig-obj doesn't seem to fix it.
 build:mingw_gcc --cxxopt=-O2
 
--- a/llvm-bazel/.bazelrc
+++ b/llvm-bazel/.bazelrc
@@ -72,11 +72,10 @@ build:generic_gcc --copt=-Werror --host_copt=-Werror
 # Windows specific flags for building with GCC. Used on top of generic_gcc
 ###############################################################################
 
+build:mingw_gcc --config=generic_gcc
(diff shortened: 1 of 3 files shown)


llvm-bazel/.bazelrc
```diff
@@ -68,6 +68,23 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
 # Use `-Werror` for GCC to make sure warnings don't slip past.
 build:generic_gcc --copt=-Werror --host_copt=-Werror
 
+###############################################################################
+# Windows specific flags for building with GCC. Used on top of generic_gcc
+###############################################################################
+
+# Other windows toolchains have these by default
+build:mingw_gcc --cxxopt=-DHAVE_UNISTD_H=1 --host_cxxopt=-DHAVE_UNISTD_H=1
```
GMNGeoffrey, line 79: Can this be added to https://github.com/google/llvm-bazel/blob/main/llvm-bazel/llvm-project-overlay/llvm/config.bzl ?

llvm-bazel/.bazelrc
```diff
@@ -68,6 +68,23 @@ build:generic_gcc --copt=-Wno-misleading-indentation --host_copt=-Wno-misleading
...
+build:mingw_gcc --linkopt=-lole32 --host_linkopt=-lole32
+build:mingw_gcc --linkopt=-luuid --host_linkopt=-luuid
+build:mingw_gcc --linkopt=-lversion --host_linkopt=-lversion
```
GMNGeoffrey: I'm guessing these are deps of particular libraries? It would be nice to not have to add them globally

llvm-bazel/.bazelrc
rubensf: I could run builds (at least partially) using clang-cl and msvc without those on `config.bzl` - it seems they're picked by default? 

Since only mingw needed them, I put them here instead. Wdyt? Or do you mean setting a bazel constraint set for when mingw is used?

(my compiler-foo-ness is subpar so bear with me if I'm missing something obvious :))

GMNGeoffrey: > Or do you mean setting a bazel constraint set for when mingw is used?

I meant this. This is ok though.

What we should really do is detect the machine capabilities and encode it that way. The current method is really just a shortcut because Bazel is not well-equipped for that.

rubensf: I tried doing what you asked, but bazel is _really_ challenging. Trying to set select conditions on anything that's not `@bazel_tools//src/conditions:os` (eg `@bazel_tools//tools/cpp:mingw`, `//:my_platform`, `//:my_constraint`) yielded either a "Your target can't be found" or "Your BUILD file can't be found" or "invalid constraint key" :/

llvm-bazel/.bazelrc
rubensf: I moved them into the specific libraries. I also tried compiling them with `msvc`, which would complain about LNK4044... I tried adding `(host_)linkopt=/IGNORE:4044` but no luck, the warnings are still there.

GMNGeoffrey: So as currently configured (with them on the specific targets), this adds LNK4044 warnings to MSVC? That's worse than having them in the bazelrc, I think. I can take a look next week about trying to detect mingw properly, but otherwise let's just leave things in the bazelrc as you had previously. @chandlerc thoughts?

rubensf: When trying to run with 
```
bazelisk build --config=msvc --compiler=msvc @llvm-project//llvm:llvm-tblgen
...
LINK : warning LNK4044: unrecognized option '/lole32'; ignored
LINK : warning LNK4044: unrecognized option '/luuid'; ignored
...
```

I _think_ these [are safe](https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-warning-lnk4044?view=msvc-160), albeit annoyingly verbose?

GMNGeoffrey: Yeah warning spam is IMO indirectly unsafe though because it makes people ignore other warnings. I think it's not fair to degrade the MSVC build experience to add support for mingw_gcc

rubensf: Done.

GMNGeoffrey (approved):

rubensf:
Also added `--compiler=mingw-gcc`

PR #178, opened 2021-03-23
Integrate LLVM at llvm/llvm-project@b24436ac96bd

ScottTodd:
No `BUILD` file changes needed for llvm/llvm-project@b24436ac96bd.


GMNGeoffrey:
Then this PR is unnecessary :-) That's why https://github.com/google/llvm-bazel/blob/main/.github/workflows/auto_integrate.yml exists. We create and tag a commit for each LLVM commit. You don't have to pick one that is on the llvm-bazel main branch. In this case https://github.com/google/llvm-bazel/releases/tag/llvm-project-b24436ac96bdf3f2c545fc85dc8af239d618c9c4 already exists

PR #180, opened 2021-03-26
Add config define line for linux aarch64.

cdleary:
Testing: with this change I was able to get the JIT workloads in the  dependent project google/xls to work on a Raspberry Pi 4.

Please do let me know if there's any other testing that you'd like to see in order to accept this PR.

### DIFF
--- a/llvm-bazel/llvm-project-overlay/llvm/config.bzl
+++ b/llvm-bazel/llvm-project-overlay/llvm/config.bzl
@@ -72,5 +72,6 @@ os_defines = select({
 llvm_config_defines = os_defines + select({
     "@bazel_tools//src/conditions:windows": native_arch_defines("X86", "x86_64-pc-win32"),
     "@bazel_tools//src/conditions:darwin": native_arch_defines("X86", "x86_64-unknown-darwin"),
+    "@bazel_tools//src/conditions:linux_aarch64": native_arch_defines("AArch64", "aarch64-unknown-linux_gnu"),
     "//conditions:default": native_arch_defines("X86", "x86_64-unknown-linux-gnu"),
 })
--- a/llvm-bazel/llvm-project-overlay/llvm/config.bzl
+++ b/llvm-bazel/llvm-project-overlay/llvm/config.bzl
@@ -72,6 +72,6 @@ os_defines = select({
 llvm_config_defines = os_defines + select({
     "@bazel_tools//src/conditions:windows": native_arch_defines("X86", "x86_64-pc-win32"),
     "@bazel_tools//src/conditions:darwin": native_arch_defines("X86", "x86_64-unknown-darwin"),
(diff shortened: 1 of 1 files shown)


llvm-bazel/llvm-project-overlay/llvm/config.bzl
```diff
@@ -72,5 +72,6 @@ os_defines = select({
 llvm_config_defines = os_defines + select({
     "@bazel_tools//src/conditions:windows": native_arch_defines("X86", "x86_64-pc-win32"),
     "@bazel_tools//src/conditions:darwin": native_arch_defines("X86", "x86_64-unknown-darwin"),
+    "@bazel_tools//src/conditions:linux_aarch64": native_arch_defines("AArch64", "aarch64-unknown-linux_gnu"),
```
chandlerc: Was the _ intended, or should this be:
```suggestion
    "@bazel_tools//src/conditions:linux_aarch64": native_arch_defines("AArch64", "aarch64-unknown-linux-gnu"),
```

cdleary:
Good eye, thanks Chandler! Sorry I cribbed that from the previous llvm_host_triple value I was using in https://github.com/google/xls/issues/286#issuecomment-782557517 -- let me check this still works with this modification and I'll let you know, RPi takes a little while to build and test...

cdleary:
Looks good with it that way as well!

Read up, just out of curiosity, to compare to what was working before `llvm_bazel` and this _also_ seems more correct! At first wondered if the triple was defined by a left-leaning three-way split, so that you could only have dashes in the last piece of the triple, but realized it's the env suffix, as documented on llvm::Triple ( https://github.com/llvm/llvm-project/blob/ed956554f96c0a37cc1f1291dbad303d3f1d495c/llvm/include/llvm/ADT/Triple.h#L24 )  -- and for why it was working before, parseOs does a startswith so whether the environment "quadruple" was provided or not probably ended up mostly the same (whether or not env was GNU). The more you know! (Random refs for archaeologists: https://github.com/llvm/llvm-project/blob/ed956554f96c0a37cc1f1291dbad303d3f1d495c/llvm/lib/Support/Triple.cpp#L555 https://github.com/llvm/llvm-project/blob/ed956554f96c0a37cc1f1291dbad303d3f1d495c/llvm/lib/Support/Triple.cpp#L513 )

GMNGeoffrey (approved):

GMNGeoffrey:
FYI, we're in the process of moving this build configuration upstream (I should add that to the README). See https://groups.google.com/g/llvm-dev/c/dcis1ebRw28/. This change will be ported over, but just wanted to give you a heads up that our aim is to move these files upstream and then turn down this repo (pending me writing down how we're going to do that :grin:)

cdleary:
> heads up that our aim is to move these files upstream and then turn down this repo

Ack, thanks for the heads up!

PR #193, opened 2021-04-27
Integrate LLVM at llvm/llvm-project@4b13b7581db5

ThomasRaoux:
[4b13b7581db5](https://github.com/llvm/llvm-project/commit/4b13b7581db5)

### DIFF
--- a/llvm-bazel/llvm-project-overlay/llvm/BUILD
+++ b/llvm-bazel/llvm-project-overlay/llvm/BUILD
@@ -3308,6 +3308,18 @@ gentbl(
     td_srcs = ["include/llvm/Option/OptParser.td"],
 )
 
+gentbl(
+    name = "WindresTableGen",
+    strip_include_prefix = "tools/llvm-rc",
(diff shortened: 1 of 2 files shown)


llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -3316,8 +3328,10 @@ cc_library(
     ]),
 )
 
-cc_binary(
-    name = "llvm-rc",
+# Simulate the upstream alias by just creating the target with two different
```
GMNGeoffrey: Aw man I created a whole starlark rule for doing exactly this that's used elsewhere. Let's not do this here. I sent a change to google internal to use `binary_alias` instead. Of course GitHub is being stupid and won't let me leave a suggested edit that crosses lines it considers unchanged :roll_eyes: , so I've pushed a commit that makes the change I'd like to see

GMNGeoffrey (approved):

PR #197, opened 2021-05-03
Adjust the hardcoded tablegen include paths to match the package's name

rrbutani:
Right now, tablegen files that (transitively) depend on tablegen files within `mlir/include` fail to build if the package name passed to `llvm_configure` is something other than `llvm-project`.

<details> <summary>For example:</summary>

```tablegen
// test.td
include "mlir/IR/OpBase.td"

def Foo : Dialect {
  let name = "foo";
  let description = [{}];
  let cppNamespace = "foo";
}
```

```starlark
# BUILD
load("@llvm//mlir:tblgen.bzl", "gentbl")

gentbl(
    name = "test",
    tbl_outs = [
        ("-gen-op-decls", "test.h.inc"),
    ],
    tblgen = "@llvm//mlir:mlir-tblgen",
    td_file = "test.td",
    td_srcs = [
        "test.td",
        "@llvm//mlir:OpBaseTdFiles"
    ],
)
```
</details>

(A full workspace example is [attached](https://github.com/google/llvm-bazel/files/6416876/example.zip).)

---

This happens because the include path for `mlir/include` that's passed to tablegen is [hardcoded](https://github.com/google/llvm-bazel/blob/7d97961847d0576226a6055009da040b6201c149/llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl#L349-L353) to `external/llvm-project` in a few places; this PR changes those snippets to use the package name passed to `llvm_configure` for this path instead.

I noticed the [TODO](https://github.com/google/llvm-bazel/blob/7d97961847d0576226a6055009da040b6201c149/llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl#L349-L350) that was present in some of the places where `llvm-project` is hardcoded; if adding `//mlir:BuiltinDialectTdFiles` and friends as dependencies in the necessary places is preferred I can try to update this PR to do that instead.

### DIFF
--- a/llvm-bazel/llvm-project-overlay/llvm/tblgen.bzl
+++ b/llvm-bazel/llvm-project-overlay/llvm/tblgen.bzl
@@ -35,6 +35,9 @@ def gentbl(
       tblgen_args: Extra arguments string to pass to the tblgen binary.
       **kwargs: Keyword arguments to pass to subsidiary cc_library() rule.
     """
+    llvm_project_workspace = Label("//:...", relative_to_caller_repository = False)
+    llvm_project_execroot_path = llvm_project_workspace.workspace_root
+
     if td_file not in td_srcs:
         td_srcs += [td_file]
     for (opts, out) in tbl_outs:
@@ -45,11 +48,13 @@ def gentbl(
             outs = [out],
             tools = [tblgen],
             message = "Generating code from table: %s" % td_file,
-            cmd = (("$(location %s) -I external/llvm-project/llvm/include " +
-                    "-I external/llvm-project/clang/include " +
+            cmd = (("$(location %s) -I %s/llvm/include " +
+                    "-I %s/clang/include " +
                     "-I $$(dirname $(location %s)) " +
                     "%s $(location %s) %s -o $@") % (
                 tblgen,
+                llvm_project_execroot_path,
+                llvm_project_execroot_path,
                 td_file,
                 opts,
                 td_file,
--- a/llvm-bazel/llvm-project-overlay/llvm/tblgen.bzl
+++ b/llvm-bazel/llvm-project-overlay/llvm/tblgen.bzl
@@ -35,8 +35,7 @@ def gentbl(
       tblgen_args: Extra arguments string to pass to the tblgen binary.
       **kwargs: Keyword arguments to pass to subsidiary cc_library() rule.
     """
-    llvm_project_workspace = Label("//:...", relative_to_caller_repository = False)
-    llvm_project_execroot_path = llvm_project_workspace.workspace_root
+    llvm_project_execroot_path = Label("//llvm:tblgen.bzl", relative_to_caller_repository = False).workspace_root
 
     if td_file not in td_srcs:
         td_srcs += [td_file]
--- a/llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
+++ b/llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
@@ -324,8 +324,7 @@ def gentbl(
       test: whether to create a shell test that invokes the tool too.
       **kwargs: Extra keyword arguments to pass to all generated rules.
     """
-    llvm_project_workspace = Label("//:...", relative_to_caller_repository = False)
-    llvm_project_execroot_path = llvm_project_workspace.workspace_root
+    llvm_project_execroot_path = Label("//llvm:tblgen.bzl", relative_to_caller_repository = False).workspace_root
 
     for (opts_string, out) in tbl_outs:
         # TODO(gcmn): The API of opts as single string is preserved for backward
--- a/llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
+++ b/llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
@@ -324,6 +324,9 @@ def gentbl(
       test: whether to create a shell test that invokes the tool too.
       **kwargs: Extra keyword arguments to pass to all generated rules.
     """
+    llvm_project_workspace = Label("//:...", relative_to_caller_repository = False)
+    llvm_project_execroot_path = llvm_project_workspace.workspace_root
+
     for (opts_string, out) in tbl_outs:
         # TODO(gcmn): The API of opts as single string is preserved for backward
         # compatibility. Change to taking a sequence.
@@ -349,7 +352,7 @@ def gentbl(
             # TODO(gcmn): Update callers to td_library and explicit includes and
             # drop this hardcoded include.
             td_includes = td_includes + [
-                "external/llvm-project/mlir/include",
+                "%s/mlir/include" % llvm_project_execroot_path,
             ],
             out = out,
             **kwargs
@@ -369,7 +372,7 @@ def gentbl(
                 # TODO(gcmn): Update callers to td_library and explicit includes
                 # and drop this hardcoded include.
                 td_includes = td_includes + [
-                    "external/llvm-project/mlir/include",
+                    "%s/mlir/include" % llvm_project_execroot_path,
                 ],
                 **kwargs
             )
--- a/llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
+++ b/llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
@@ -324,7 +324,7 @@ def gentbl(
       test: whether to create a shell test that invokes the tool too.
       **kwargs: Extra keyword arguments to pass to all generated rules.
     """
-    llvm_project_execroot_path = Label("//llvm:tblgen.bzl", relative_to_caller_repository = False).workspace_root
+    llvm_project_execroot_path = Label("//mlir:tblgen.bzl", relative_to_caller_repository = False).workspace_root
 
     for (opts_string, out) in tbl_outs:
         # TODO(gcmn): The API of opts as single string is preserved for backward


llvm-bazel/llvm-project-overlay/llvm/tblgen.bzl
```diff
@@ -35,6 +35,9 @@ def gentbl(
       tblgen_args: Extra arguments string to pass to the tblgen binary.
       **kwargs: Keyword arguments to pass to subsidiary cc_library() rule.
     """
+    llvm_project_workspace = Label("//:...", relative_to_caller_repository = False)
+    llvm_project_execroot_path = llvm_project_workspace.workspace_root
```
GMNGeoffrey: I think this isn't really a well-formed label. What about using this file itself. We know that exists :-)

```suggestion
    llvm_project_execroot_path = Label("//llvm:tblgen.bzl", relative_to_caller_repository = False).workspace_root
```

GMNGeoffrey (commented):
Yeah the current hardcoding is awful. That's half the reason I rewrote gentbl as a rule. Adding the correct dependencies is indeed the preferred solution if it is sufficient. It may not be though, since LLVM is still using the old rule. The plan is to delete these hardcoded includes entirely and merge the tablegen rules. I haven't finished the cleanup though. This change is probably an improvement either way. I didn't know about that trick for getting the workspace root in a macro.

Can you include the error messages you see?

rrbutani:
> I didn't know about that trick for getting the workspace root in a macro.

Yeah... it feels like there should be a better way to do it but I haven't found one. 😕 

> Can you include the error messages you see?

Sure, my bad.

For `bazel build //:test.h.inc` I get:
```
test.td:1:9: error: Could not find include file 'mlir/IR/OpBase.td'
include "mlir/IR/OpBase.td"
        ^
test.td:1:9: error: Unexpected token at top level
include "mlir/IR/OpBase.td"
        ^
```

Running with `--verbose_failures` shows that the include paths being passed to tablegen point to `external/llvm-project/mlir/include` even when the package is named something else.

llvm-bazel/llvm-project-overlay/llvm/tblgen.bzl
rrbutani: I don't think it _actually_ matters since we never use the label itself anywhere but sure 👍

GMNGeoffrey: Yeah probably not :-D Using `...` makes it seem like it's somehow a package expression, but it's really referencing a nonexistent file called "...". I'd be fine with something obviously non-existent too like "//:foobar" or "//:dummy"

GMNGeoffrey:
> > I didn't know about that trick for getting the workspace root in a macro.
> 
> Yeah... it feels like there should be a better way to do it but I haven't found one. 
> 
> > Can you include the error messages you see?
> 
> Sure, my bad.
> 
> For `bazel build //:test.h.inc` I get:
> 
> ```
> test.td:1:9: error: Could not find include file 'mlir/IR/OpBase.td'
> include "mlir/IR/OpBase.td"
>         ^
> test.td:1:9: error: Unexpected token at top level
> include "mlir/IR/OpBase.td"
>         ^
> ```


Ah yeah this mean we have missing dependencies hidden by the hardcoding :cry: The hardcoding is still in there because dependent projects have to also be using td_library to get the benefit, so I didn't delete it yet. If you wouldn't mind adding the correct dependencies as well, that would be great, but for now, there's no reason to be privileging dependent projects that call this repo specifically "llvm-project" and this avoids depending on the "external" thing which I think is a Bazel implementation detail.

> Running with `--verbose_failures` shows that the include paths being passed to tablegen point to `external/llvm-project/mlir/include` even when the package is named something else.

Yeah it's not surprising they're getting the hardcoded include. They should also be depending on the package that contains OpBase.td though and then they'll get the correct include path as well.

llvm-bazel/llvm-project-overlay/mlir/tblgen.bzl
```diff
@@ -324,6 +324,8 @@ def gentbl(
       test: whether to create a shell test that invokes the tool too.
       **kwargs: Extra keyword arguments to pass to all generated rules.
     """
+    llvm_project_execroot_path = Label("//llvm:tblgen.bzl", relative_to_caller_repository = False).workspace_root
```
GMNGeoffrey: Nit: let's use this file's name instead

```suggestion
    llvm_project_execroot_path = Label("//mlir:tblgen.bzl", relative_to_caller_repository = False).workspace_root
```

GMNGeoffrey (approved):

rrbutani:
> Yeah it's not surprising they're getting the hardcoded include. They should also be depending on the package that contains OpBase.td though and then they'll get the correct include path as well.

Oh whoops 🤦. I put `"@llvm//mlir:OpBaseTdFiles"` under `td_srcs` instead of `deps` which is why I was getting that error in the example I posted.

Ignoring that, I was still getting errors for some transitive tablegen dependencies; I double checked and these were the dependency changes I had to make to the mlir BUILD file:
```diff
+++ b/mlir.BUILD
@@ -125,6 +125,7 @@ gentbl(
         "include/mlir/IR/BuiltinDialect.td",
         ":OpBaseTdFiles",
     ],
+    deps = [":BuiltinDialectTdFiles"],
 )
 
 gentbl(
@@ -3224,7 +3225,7 @@ gentbl(
     ],
     tblgen = ":mlir-tblgen",
     td_file = "include/mlir/Interfaces/DataLayoutInterfaces.td",
-    td_srcs = [":OpBaseTdFiles"],
+    deps = [":OpBaseTdFiles"],
 )
 
 gentbl(
@@ -5725,9 +5726,7 @@ gentbl(
     ],
     tblgen = ":mlir-tblgen",
     td_file = "include/mlir/Dialect/DLTI/DLTIBase.td",
-    td_srcs = [
-        ":OpBaseTdFiles",
-    ],
+    deps = [":OpBaseTdFiles"],
 )
 
 cc_library(
```


<details> <summary>With more context:</summary>

```diff
@@ -108,40 +108,41 @@ gentbl(
 gentbl(
     name = "BuiltinLocationAttributesIncGen",
     strip_include_prefix = "include",
     tbl_outs = [
         (
             "--gen-attrdef-decls",
             "include/mlir/IR/BuiltinLocationAttributes.h.inc",
         ),
         (
             "--gen-attrdef-defs",
             "include/mlir/IR/BuiltinLocationAttributes.cpp.inc",
         ),
     ],
     tblgen = ":mlir-tblgen",
     td_file = "include/mlir/IR/BuiltinLocationAttributes.td",
     td_srcs = [
         "include/mlir/IR/BuiltinLocationAttributes.td",
         "include/mlir/IR/BuiltinDialect.td",
         ":OpBaseTdFiles",
     ],
+    deps = [":BuiltinDialectTdFiles"],
 )
 
 gentbl(
     name = "BuiltinOpsIncGen",
     strip_include_prefix = "include",
     tbl_outs = [
         (
             "-gen-op-decls",
             "include/mlir/IR/BuiltinOps.h.inc",
         ),
         (
             "-gen-op-defs",
             "include/mlir/IR/BuiltinOps.cpp.inc",
         ),
     ],
     tblgen = ":mlir-tblgen",
     td_file = "include/mlir/IR/BuiltinOps.td",
     deps = [":BuiltinDialectTdFiles"],
 )
 
@@ -3207,41 +3208,41 @@ gentbl(
         ),
         (
             "-gen-op-interface-decls",
             "include/mlir/Interfaces/DataLayoutOpInterface.h.inc",
         ),
         (
             "-gen-op-interface-defs",
             "include/mlir/Interfaces/DataLayoutOpInterface.cpp.inc",
         ),
         (
             "-gen-type-interface-decls",
             "include/mlir/Interfaces/DataLayoutTypeInterface.h.inc",
         ),
         (
             "-gen-type-interface-defs",
             "include/mlir/Interfaces/DataLayoutTypeInterface.cpp.inc",
         ),
     ],
     tblgen = ":mlir-tblgen",
     td_file = "include/mlir/Interfaces/DataLayoutInterfaces.td",
-    td_srcs = [":OpBaseTdFiles"],
+    deps = [":OpBaseTdFiles"],
 )
 
 gentbl(
     name = "LoopLikeInterfaceIncGen",
     strip_include_prefix = "include",
     tbl_outs = [
         (
             "-gen-op-interface-decls",
             "include/mlir/Interfaces/LoopLikeInterface.h.inc",
         ),
         (
             "-gen-op-interface-defs",
             "include/mlir/Interfaces/LoopLikeInterface.cpp.inc",
         ),
     ],
     tblgen = ":mlir-tblgen",
     td_file = "include/mlir/Interfaces/LoopLikeInterface.td",
     deps = [":LoopLikeInterfaceTdFiles"],
 )
 
@@ -5708,43 +5709,41 @@ td_library(
     name = "DLTIDialectTdFiles",
     srcs = [
         "include/mlir/Dialect/DLTI/DLTI.td",
         "include/mlir/Dialect/DLTI/DLTIBase.td",
     ],
     includes = ["include"],
     deps = [":OpBaseTdFiles"],
 )
 
 gentbl(
     name = "DLTIBaseIncGen",
     strip_include_prefix = "include",
     tbl_outs = [
         (
             "-gen-dialect-decls -dialect=dlti",
             "include/mlir/Dialect/DLTI/DLTIDialect.h.inc",
         ),
     ],
     tblgen = ":mlir-tblgen",
     td_file = "include/mlir/Dialect/DLTI/DLTIBase.td",
-    td_srcs = [
-        ":OpBaseTdFiles",
-    ],
+    deps = [":OpBaseTdFiles"],
 )
 
 cc_library(
     name = "DLTIDialect",
     srcs = glob(["lib/Dialect/DLTI/*.cpp"]),
     hdrs = glob(["include/mlir/Dialect/DLTI/*.h"]),
     includes = ["include"],
     deps = [
         ":DLTIBaseIncGen",
         ":DataLayoutInterfaces",
         ":IR",
         "//llvm:Support",
     ],
 )
 
 gentbl(
     name = "ReducerIncGen",
     strip_include_prefix = "include",
     tbl_outs = [
         (

```
</details>

(against `llvm-project-2f69975683f5b6ea7df79f335f96b889a4bddecd`)


Some LLVM tablegen targets had some other errors that I didn't try to figure out.

---

Anyways, thanks for the super speedy response! 😃

GMNGeoffrey:
Fur sure. If you want to send that patch as a PR, that would be great :smile: otherwise I can patch it in

PR #198, opened 2021-05-03
Add in some missing tablegen deps

rrbutani:
A follow up to #197.


rrbutani:
There are some missing ones on the LLVM side too; I gave up when I saw the list comprehension in the LLVM BUILD file but I'll try again.

GMNGeoffrey:
> There are some missing ones on the LLVM side too; I gave up when I saw the list comprehension in the LLVM BUILD file but I'll try again.

LLVM is still using the old macro that doesn't support dependencies. So don't put in that effort!

GMNGeoffrey (approved):

GMNGeoffrey:
Thanks!

PR #200, opened 2021-05-08
Further cleanup of llvm/BUILD

GMNGeoffrey:
This is a port of diffs against our internal copy of these build files, which have inevitably
drifted apart. Mostly adding missing targets and deps.

The only potentially controversial change is flipping the default for
LLVM_ENABLE_PERF (from disabled to enabled).


llvm-bazel/llvm-project-overlay/llvm/BUILD
```diff
@@ -841,7 +843,7 @@ cc_library(
             "lib/Analysis/*.def",
         ],
         exclude = [
-            # TODO: Add these files to the build.
+            # TODO(mtrofin): Add these files to the build.
```
chandlerc, line 846: Leaving these out was intentional -- I don't think this adds value in the open source version.

If we want to add LLVM bug links, that would work, but otherwise I'd just remove internal usernames.

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
```diff
@@ -68,7 +68,7 @@
 #define LLVM_USE_OPROFILE 0
 
 /* Define if we have the perf JIT-support library */
-#define LLVM_USE_PERF 0
+#define LLVM_USE_PERF 1
```
chandlerc: I've hit distros where this isn't available... is it important to change this default?

We could move this to the configured macro set like we do for some other libraries where we support both using and not using them.

chandlerc (changes requested):
Cool, mostly minor tweaks.

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: It's a github username too. We have TODO(gcmn) and TODO(chandlerc) floating around here, so this seemed ok

GMNGeoffrey: (they were originally internal bug links, which I agree was not useful :stuck_out_tongue: )

rupprecht (approved):
Nothing on top of what Chandler said.

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
GMNGeoffrey: Yeah I figured this would be the most controversial thing. We can move it to something more configurable. One of the wrinkles is that we have to actually not add PerfJITEventListener.cpp to the srcs if this set off, which is a bit annoying, because of a redefinition of createPerfJITEventListener:

> ERROR: /usr/local/google/home/gcmn/.cache/bazel/_bazel_gcmn/1655d0ec87ba8569e84422cbb0aeec46/external/llvm-project/llvm/BUILD:1996:11: C++ compilation of rule
 '@llvm-project//llvm:ExecutionEngine' failed (Exit 1)                                                                                                        
external/llvm-project/llvm/lib/ExecutionEngine/PerfJITEvents/PerfJITEventListener.cpp:495:37: error: redefinition of 'createPerfJITEventListener'             
JITEventListener *JITEventListener::createPerfJITEventListener() {                                                                                            
                                    ^                                                                                                                         
external/llvm-project/llvm/include/llvm/ExecutionEngine/JITEventListener.h:102:28: note: previous definition is here                                          
  static JITEventListener *createPerfJITEventListener()  

In CMake, this is gated with an add_subdirectory. We should probably break up the ExecutionEngine target, but I have this sneaking suspicion it is the way it is because of bad layering. Thoughts?

GMNGeoffrey: Also looks like this is strictly only supported on linux, so we should at least gate it on that

llvm-bazel/llvm-project-overlay/llvm/BUILD
chandlerc: Ok.... Not sure how we're going to keep that in sync, but sure.

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
chandlerc: Ooof.

I don't see a good way of doing this other than breaking up `ExecutionEngine` so that we can avoid building the code when this is missing.

I'd suggest just keeping this internal until that refactoring can land.

I'm hopeful the layering isn't too bad though -- my guess is we just hadn't gotten to the point of breaking it up.

llvm-bazel/llvm-project-overlay/llvm/config.bzl
```diff
@@ -41,6 +41,7 @@ linux_defines = posix_defines + [
     "HAVE_POSIX_FALLOCATE=1",
     "HAVE_SBRK=1",
     "HAVE_STRUCT_STAT_ST_MTIM_TV_NSEC=1",
+    "LLVM_USE_PERF=1",
```
chandlerc: I'm nervous about even doing this in upstream. I think this really needs to be detected on the system by Bazel before we enable it.

I'd just keep all of this stuff disabled until the build files land fully, and then work on enabling it in the upstream build because I think we'll need to spend some time getting the conditions correct.

llvm-bazel/llvm-project-overlay/llvm/BUILD
GMNGeoffrey: Well we will soon be deriving the internal version from upstream rather than the other way around. I'm writing the scripting right now and this seemed like a fine thing to have upstream and not something worth adding special handling to add on import :grin:

llvm-bazel/llvm-project-overlay/llvm/include/llvm/Config/llvm-config.h
GMNGeoffrey: Ok I'll do that.

llvm-bazel/llvm-project-overlay/llvm/config.bzl
GMNGeoffrey: Done

chandlerc (approved):
LGTM!

PR #202, opened 2021-05-11
Drop visibility from MLIR build

GMNGeoffrey:
Visibility is a useful restriction within a repo, but across repos it really doesn't work very
well. These visibility markers right now mostly do nothing anyway. Just drop it entirely and
make everything public. I have some vague ideas of how to make Bazel visibility work by
turning it into hooks that a dependent repo can choose to implement, but it's not high on
the list of priorities.

### DIFF
--- a/llvm-bazel/llvm-project-overlay/mlir/BUILD
+++ b/llvm-bazel/llvm-project-overlay/mlir/BUILD
@@ -10,18 +10,8 @@ load(":linalggen.bzl", "genlinalg")
 load(":build_defs.bzl", "cc_headers_only", "if_cuda_available")
 
 package(
-    default_visibility = [":friends"],
+    default_visibility = ["//visibility:public"],
     licenses = ["notice"],
-)
-
-package_group(
-    name = "subpackages",
-    packages = ["//mlir/..."],
-)
-
-package_group(
-    name = "friends",
-    packages = ["//..."],
 )
 
 exports_files([
@@ -33,9 +23,6 @@ exports_files([
 filegroup(
     name = "c_headers",
     srcs = glob(["include/mlir-c/**/*"]),  # <== i.e. match the entire tree
-    visibility = [
-        ":friends",
-    ],
 )
 
 [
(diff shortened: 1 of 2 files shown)


chandlerc (approved):
I'm fine with this but I wonder if it makes sense to keep a more basic public vs. private where really only the complete internal details are private?

Somewhat up to you and/or the other MLIR folks.

Approving for now so you're not blocked if this is the path forward.

GMNGeoffrey:
Yeah I think adding such a thing might make sense, but it's not really what we have now, which is basically just public visibility by a more complicated mechanism. The issue to me is that restricting visibility makes a lot of sense in google's internal source, where the burden of maintaining users of an API generally falls on the API author. You really don't want to give people visibility into your internal stuff. But in OSS, if someone uses something internal and you change it, that's generally on them to fix. And consequently, if they want to make a considered decision to use stuff marked internal anyway, then that's up to them and their priorities (maybe they're experimenting with some contribution that would be upstreamed leverage internal stuff, for instance). MLIR (and LLVM as well, though less so today) routinely makes breaking changes far more dramatic, so I don't think this is going to stand out. I think the nice thing to do is to have visibility tagged as a guide for users, so they don't *accidentally* use internal things and then get broken, but visibility doesn't scale well to that because there isn't a way to override it from outside the package. The only option today is the very heavy hammer of just disabling it entirely with a Bazel flag, which means you wouldn't get the benefits from any project. The other thing is that it doesn't let someone make visibility more restrictive. Like say I want to ensure that only my "llvm integration" can use llvm stuff. There's no way to do that if the LLVM public APIs have been marked with public visibility. So my idea would be to have basically a secondary repo of `@llvm-project-visibility` that all the visibility rules reference and then provide a default implementation of that that keeps things nice, but also allow someone to set up their own in their WORKSPACE. But it's a fair amount of fiddling for not that large a benefit and I recall that the visibility builtins don't act like normal targets in a number of ways, so I'm not sure whether it would even work.

PR #213, opened 2021-05-20
Add helper macros for configuring external llvm:Support deps

GMNGeoffrey:
This bundles some basic configurations for the llvm:Support deps that
must be configured (including as disabled).

This is in response to feedback on the patch to upstream:
https://reviews.llvm.org/D90352#inline-972755

Tested: I confirmed that this works with IREE's current usage.


chandlerc (approved):
LGTM

GMNGeoffrey:
GitHub's merge algorithm sucks...

PR #215, opened 2021-05-23
Allow configuring the set of LLVM targets built.

chandlerc:
For library users that only need specific targets built, this is
particularly significant both to speed up builds and also avoid
embedding large amounts of unnecessary code.

This somewhat surfaced that the repository rule being the overlay
creation doesn't really make sense, and so I've folded that into
a single repository rule where it is reasonable to add LLVM-specific
logic to fully customize things. Happy to use a different structure if
folks have preferences.

I also know that all of this is on the way upstream, but I'd like to use
this ASAP, and didn't want to necessarily gate on the upstream patches
landing if that's OK.


llvm-bazel/llvm-project-overlay/llvm/BUILD.bazel
```diff
@@ -513,6 +501,8 @@ gentbl(
     ]),
 )
 
+# Note that the intrinsics are not currently set up so they can be pruneed for
```
GMNGeoffrey: Nit: typo

```suggestion
# Note that the intrinsics are not currently set up so they can be pruned for
```

llvm-bazel/llvm-project-overlay/llvm/BUILD.bazel
```diff
@@ -68,7 +55,7 @@ llvm_target_asm_parsers = [
     "Sparc",
     "WebAssembly",
     "X86",
-]
+] if t in llvm_targets]
```
GMNGeoffrey, line 58: I wonder if it would make more sense to use this configuration to limit what gets pulled in to common "AllTargetsFoo" targets rather than removing the targets entirely. Then building or depending on a specific target directly doesn't require reconfiguring the repository, but it still limits the targets that just get pulled in automatically. For instance, there are targets in MLIR that depend specifically on NVPTX and I might want to be able to build those while still not having mlir-opt include NVPTX. Maybe even have configurations for both since "I want to guarantee that nothing I use depends on NVPTX" is also a good goal. Something like "default targets" and "available targets"

llvm-bazel/configure.bzl
```diff
@@ -4,10 +4,113 @@
...
+      `True` if `path` is an absolute path.
+    """
+    return path.startswith("/") or (len(path) > 2 and path[1] == ":")
+
+def _join_path(a, b):
+    if _is_absolute(b):
+        return b
+    return str(a) + "/" + str(b)
+
+def _overlay_directories(repository_ctx):
```
GMNGeoffrey, line 40: I had this as a separate repo rule mostly for conceptual organization and because I thought it was a useful technology on its own. That doesn't mean it needs to be in a separate file though, unless we actually do find another use for it and want to move it somewhere else.

GMNGeoffrey (commented):
Regarding making this change before upstreaming, there isn't a hard block. It's just more maintenance work to port changes back to our internal build as well to keep them in sync.

llvm-bazel/llvm-project-overlay/llvm/BUILD.bazel
chandlerc: It's a bit tricky...

So, if we do this, it'll let code build against the target, but using it will be a bit "exciting"....

Imagine you have a test `a_test` that depends on a library `a` which in turn depends on library `b`, and it is `b` that depends on NVPTX. The default initialize-all helper routines that `a_test` might use won't actually initialize the NVPTX target, and so `a_test` would have to add custom initialization to explicitly initialize NVPTX which seems ... surprising.

Weirdly, most of the time, things will "just work" because there are a bunch of redundant attempts to initialize targets, but there may be strange and mysterious failures here.

Anyways, I'm happy to go either way -- I mostly want to surface that this isn't a trivial pattern to follow. Let me know which you prefer.

llvm-bazel/configure.bzl
chandlerc: Yeah, it was tricky to have in a separate repo rule because then any LLVM-specific configuration would need to be under a different named repo.

Can always share the starlark *code* here if that becomes useful while keeping a simple repo rule structure?

chandlerc:
> Regarding making this change before upstreaming, there isn't a hard block. It's just more maintenance work to port changes back to our internal build as well to keep them in sync.

If possible, I'd love to land it sooner just so that I can use it to speed up some builds... but ultimately, up to you. =] Don't want to interfere with the upstreaming.

llvm-bazel/configure.bzl
GMNGeoffrey: Yeah totally understand what happened here. I think have a separate function that takes a repository_ctx (as you have) is the best thing to do here and we can always break that function out if we have a need.

llvm-bazel/llvm-project-overlay/llvm/BUILD.bazel
GMNGeoffrey: That sounds like a good reason to at least hold off until we actually have a use case. Fully disabling the targets (like not even configuring them) is the way the cmake build works, right? The issue with the current setup is going to be that MLIR has a hard dependency on NVPTX. I guess that's ok because it's also true today though. Ugh cuda...

GMNGeoffrey (approved):

llvm-bazel/configure.bzl
chandlerc: Awesome, and thanks for the review!

PR #218, opened 2021-05-28
Integrate LLVM at llvm/llvm-project@6359842bc088

GMNGeoffrey:
[6359842bc088](https://github.com/llvm/llvm-project/commit/6359842bc088)

No change actually *needed* for this, but allows removing a set of
headers from support that are no longer necessary. While I was here,
fixed a TODO pointing at an internal bug number.


chandlerc (approved):
LGTM

GMNGeoffrey:
> LGTM

Thanks!

PR #236, opened 2021-07-22
Change external_zlib attribute to string

mmcloughlin:
When using `llvm_zlib_external` rule with `external_zlib` attribute set to a label referring to the main repository, like `@//third_party/zlib`, it will be replaced with `//third_party/zlib` after template substitution. This will then attempt to find `//third_party/zlib` within the local repository `@llvm_zlib//third_party/zlib`, which does not exist, rather than the intended reference back to the main repository. The issue appears to be that the conversion of `Label` type to string with `str(repository_ctx.attr.external_zlib)`, which is causing the main repository qualifier to be lost.

This PR fixes the issue by changing the `external_zlib` attribute to `attr.string` type rather than `attr.label`.

### DIFF
--- a/llvm-bazel/zlib.bzl
+++ b/llvm-bazel/zlib.bzl
@@ -24,7 +24,7 @@ def _llvm_zlib_external_impl(repository_ctx):
         "BUILD",
         repository_ctx.attr._external_build_template,
         substitutions = {
-            "@external_zlib_repo//:zlib_rule": str(repository_ctx.attr.external_zlib),
+            "@external_zlib_repo//:zlib_rule": repository_ctx.attr.external_zlib,
         },
         executable = False,
     )
@@ -36,7 +36,7 @@ llvm_zlib_external = repository_rule(
             default = Label("//deps_impl:zlib_external.BUILD"),
             allow_single_file = True,
         ),
-        "external_zlib": attr.label(
+        "external_zlib": attr.string(
             doc = "The dependency that should be used for the external zlib library.",
             mandatory = True,
         ),


mmcloughlin:
For reference, the error I currently see:

```
$ bazel build @llvm-project//clang:tooling
INFO: Invocation ID: 58a0cb73-da65-4c2c-bf70-1f34409f8692
ERROR: /home/mbm/.cache/bazel/_bazel_mbm/a66b6aa2b99f3171a7f938ac114f0763/external/llvm_zlib/BUILD:6:11: no such package '@llvm_zlib//third_party/zlib': BUILD file not found in directory 'third_party/zlib' of external repository @llvm_zlib. Add a BUILD file to a directory to mark it as a package. and referenced by '@llvm_zlib//:zlib'
ERROR: Analysis of target '@llvm-project//clang:tooling' failed; build aborted: Analysis failed
INFO: Elapsed time: 1.295s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (3 packages loaded, 124 targets configured)
```

GMNGeoffrey:
Thanks for the fix. These build files have no moved upstream to https://github.com/llvm/llvm-project/tree/main/utils/bazel. Are you still using them from this repository? I'm happy to merge this PR here if you're still pointing at this one, but would you consider also sending a patch upstream? I can send it myself, but wanted you to get the credit if you want it :-)

I'm not planning to make further updates to this repository myself, but I'm happy to continue merging PRs for a bit to keep things working until folks switch over. I'll update the README to this effect

GMNGeoffrey:
Here's the readme update: https://github.com/google/llvm-bazel/pull/237

llvm-bazel/zlib.bzl
```diff
@@ -36,7 +36,7 @@ llvm_zlib_external = repository_rule(
             default = Label("//deps_impl:zlib_external.BUILD"),
             allow_single_file = True,
         ),
-        "external_zlib": attr.label(
+        "external_zlib": attr.string(
```
GMNGeoffrey, line 39: I feel like Bazel is doing the wrong thing here when converting the label to a string. It shouldn't be dropping the leading "@" if that is load-bearing

GMNGeoffrey (approved):

mmcloughlin:
Oh that's excellent news, congrats on getting this upstreamed. Yeah I'll look at switching our install to the in-tree bazel rules. I'll also send this patch upstream.

llvm-bazel/zlib.bzl
mmcloughlin: Right. Agreed on this, maybe I'll take a look at the Label implementation and see what's going on. Either there's a bug or you're supposed to use the Label.relative() method somehow. But after a quick look at docs I couldn't figure it out and thought string would be good enough.

GMNGeoffrey: Yeah I did the same and couldn't figure out what the "right" way. Can you file an issue/start a discussion?

mmcloughlin: The function that does what you want appears to be here:

https://github.com/bazelbuild/bazel/blob/227da61a0a522e5c3e3cd3f05e2e2d3a0a0c1b00/src/main/java/com/google/devtools/build/lib/cmdline/Label.java#L438-L444

But this isn't exposed through the starlark bindings. I'll file a question with bazel to ask what the right solution is.

mmcloughlin: Filed bazelbuild/bazel#13731.

mmcloughlin:
Upstream diff: https://reviews.llvm.org/D106606.

